PR #364, opened 2022-08-12
feat: implement new 'internal/metrics/v2' package

hbagdi:
'internal/metrics' package provides some abstraction over metrics but it
    doesn't provide enough control or abstraction. Specifically:
    - no support for subsystem in naming (unless enforced manually
    - no support for buckets in histograms

    The new package introduces a few functions that wrap Prometheus details.
    This is done to ease introduction of new metrics packages, while
    providing an implementation that takes full advantages of
    Prometheus-specific features.
    Support for other metrics backends has been dropped and may be
    introduced in the future based on demand.

    With the introduction of this new package, 'internal/metrics' is
    considered deprecated and will be removed in a following patch.


internal/metrics/v2/counter.go
```diff
@@ -0,0 +1,44 @@
...
+// Counter is a metrics gauge.
+// Please consult Prometheus documentation for a detailed understanding.
+type Counter interface {
+	// Inc increments the gauge by 1. Use Add to increment it by arbitrary
+	// non-negative values.
+	Inc(...Label)
+	// Add adds the given value to the gauge. It panics if the value is <
+	// 0.
+	Add(float64, ...Label)
+}
```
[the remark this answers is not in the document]
omegabytes: Not directly. Once we change the Counter type to [prometheus.CounterVec](https://pkg.go.dev/github.com/prometheus/client_golang/prometheus#CounterVec), we get access to [a reset method](https://pkg.go.dev/github.com/prometheus/client_golang/prometheus#MetricVec.Reset) that's implemented on [MetricVet.](https://pkg.go.dev/github.com/prometheus/client_golang/prometheus#MetricVec)

internal/metrics/v2/gauge.go
```diff
@@ -0,0 +1,57 @@
...
+
+// Gauge is a metrics gauge.
+// Please consult Prometheus documentation for a detailed understanding.
+type Gauge interface {
+	// Set sets the Gauge to an arbitrary value.
+	Set(float64, ...Label)
+	// Inc increments the Gauge by 1. Use Add to increment it by arbitrary
+	// values.
+	Inc(...Label)
+	// Dec decrements the Gauge by 1. Use Sub to decrement it by arbitrary
```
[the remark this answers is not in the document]
omegabytes: [Also missing](https://pkg.go.dev/github.com/prometheus/client_golang/prometheus#Gauge): `SetToCurrentTime()`. Is that something we want?

internal/metrics/v2/histogram.go
```diff
@@ -0,0 +1,59 @@
...
+	// Metrics with the same fully-qualified name must have the same Help
+	// string.
+	Help string
+
+	// Buckets defines the buckets into which observations are counted. Each
+	// element in the slice is the upper inclusive bound of a bucket. The
+	// values must be sorted in strictly increasing order. There is no need
+	// to add a highest bucket with +Inf bound, it will be added
+	// implicitly. The default value is DefBuckets.
+	Buckets []float64
```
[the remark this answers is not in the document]
omegabytes: Fixed [here](https://github.com/Kong/koko/pull/364/commits/8a166a58e66fbad72d1c23d45e00efa989061646#diff-17aa84fbeb059b16f4bb471e34eb2248d71bc48be5bc92ce64234b9429594148)

internal/metrics/v2/counter.go
```diff
@@ -0,0 +1,44 @@
...
+
+type CounterOpts Opts
+
+func NewCounter(opts CounterOpts) Counter {
+	return &prometheusCounter{
+		counter: promauto.NewCounterVec(prometheus.CounterOpts{
```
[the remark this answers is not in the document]
omegabytes: Added a registry arg for all metrics types and implemented `promauto.With(registry)` in the constructors. See https://pkg.go.dev/github.com/prometheus/client_golang/prometheus/promauto#pkg-overview

internal/config/config.go
```diff
@@ -5,9 +5,35 @@ import (
...
+`)
+	defaultConfig Config
+)
+
+func init() {
+	err := yaml.Unmarshal(defaultConfigYAML, &defaultConfig)
+	if err != nil {
+		panic(fmt.Errorf("failed to decode default config: %v", err))
+	}
+}
```
omegabytes, line 35: @hbagdi in order to properly implement this, I need to make updates to some (all?) of the types by including `omitempty` so that zero-values don't overwrite config defaults. Is that the intended expected behavior?

omegabytes: Are we adding this default config only for testing purposes?

omegabytes: Nevermind, this was added mistakenly during a rebase. I've resolved.

internal/test/run/run.go
```diff
@@ -52,6 +52,7 @@ func WithDPAuthMode(dpAuthMode cmd.DPAuthMode) ServerConfigOpt {
 }
 
 func Koko(t *testing.T, options ...ServerConfigOpt) func() {
+	util.RegisterSchemasFromFS()
```
omegabytes, line 55: Required as of #444

internal/metrics/v2/histogram_test.go
```diff
@@ -0,0 +1,176 @@
...
+	}
+	for _, test := range tests {
+		t.Run(test.name, func(t *testing.T) {
+			h := NewHistogram(test.fields.registry, test.fields.opts)
+			for _, val := range test.args.floatVals {
+				h.Observe(val, test.args.label...)
+			}
+			family, err := test.fields.registry.Gather()
+			require.NoError(t, err)
+			histogram := family[0].Metric[0].Histogram
```
tjasko, line 171: As a sanity check, we should assert the length of `family` & `family[0].Metric` (and in all the other places).

internal/metrics/v2/gauge.go
```diff
@@ -0,0 +1,65 @@
...
+		gauge: promauto.With(registry).
+			NewGaugeVec(prometheus.GaugeOpts{
+				Namespace: namespace,
+				Subsystem: opts.Subsystem,
+				Name:      opts.Name,
+				Help:      opts.Help,
+			},
+				opts.LabelNames,
+			),
+	}
```
tjasko, line 40: nit: The formatting here (and in the other places) is really bizarre:
```suggestion
	return &prometheusGauge{
		gauge: promauto.With(registry).NewGaugeVec(
			prometheus.GaugeOpts{
				Namespace: namespace,
				Subsystem: opts.Subsystem,
				Name:      opts.Name,
				Help:      opts.Help,
			},
			opts.LabelNames,
		),
	}
```

internal/metrics/v2/gauge_test.go
```diff
@@ -0,0 +1,447 @@
...
+		},
+	}
+	for _, test := range tests {
+		t.Run(test.name, func(t *testing.T) {
+			g := NewGauge(test.fields.registry, test.fields.opts)
+			g.Sub(test.args.floatVal, test.args.label...)
+			family, err := test.fields.registry.Gather()
+			require.NoError(t, err)
+			require.Equal(t, test.expect, family[0].Metric[0].Gauge.GetValue())
+			require.Equal(t, len(test.fields.opts.LabelNames), len(family[0].Metric[0].GetLabel()))
```
tjasko, line 444: `requre.Len()` exists for this (and the other places where this is being done) 😄

internal/metrics/v2/gauge.go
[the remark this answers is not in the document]
tjasko: Can always implement it later, but yeah, it could be useful.

internal/server/kong/ws/config_service.go
```diff
@@ -73,6 +74,16 @@ func (c *configService) PingCP(
 			Value: "wrpc",
 		},
 	)
+	dpPingCounter.Inc(
```
tjasko, line 77: Can we move these new metrics to another commit, as they're not related to these changes?

internal/server/kong/ws/manager.go
```diff
@@ -22,9 +22,11 @@ import (
 	relay "github.com/kong/koko/internal/gen/grpc/kong/relay/service/v1"
 	grpcKongUtil "github.com/kong/koko/internal/gen/grpc/kong/util/v1"
 	"github.com/kong/koko/internal/metrics"
+	metricsV2 "github.com/kong/koko/internal/metrics/v2"
```
tjasko, line 25: nit: Everywhere else the import is called `metricsv2`.

internal/metrics/v2/counter_test.go
```diff
@@ -0,0 +1,198 @@
...
+		name        string
+		fields      fields
+		args        args
+		expect      float64
+		shouldPanic bool
+	}{
+		{
+			name: "should add",
+			fields: fields{
+				registry: prometheus.NewRegistry(),
```
tjasko, line 64: nit: In these tests (not just this file, but all of 'em), we can just default the registry in for loop below.

internal/metrics/v2/counter_test.go
```diff
@@ -0,0 +1,198 @@
...
+			c := NewCounter(test.fields.registry, test.fields.opts)
+			if test.shouldPanic {
+				require.Panics(t, func() {
+					c.Add(test.args.floatVal, test.args.label...)
+				})
+			} else {
+				c.Add(test.args.floatVal, test.args.label...)
+				family, err := test.fields.registry.Gather()
+				require.NoError(t, err)
+				require.Equal(t, test.expect, family[0].Metric[0].Counter.GetValue())
```
tjasko, line 139: In these tests, can we also assert that `MetricFamily.Name`, `MetricFamily.Help`, & `MetricFamily.Type` was set properly?

tjasko (commented):
Just a few minor comments, thanks for pushing this through 😄

internal/metrics/v2/gauge.go
omegabytes: interesting that it made it through gofmt, I'll double check

koko.yaml
```diff
@@ -66,4 +66,6 @@ database:
 control_server:
   tls_cert_path: cluster.crt
   tls_key_path: cluster.key
+metrics:
+  enabled: true
```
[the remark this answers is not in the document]
omegabytes: Saw that but figured I'd ask in case I'm missing something

internal/metrics/v2/gauge.go
omegabytes: ah looks like I already addressed that ;-)

internal/server/kong/ws/config_service.go
omegabytes: done [630b8de](https://github.com/Kong/koko/pull/364/commits/630b8debe9b656d21379d237ac8f3989297d1bc4)

internal/metrics/v2/counter_test.go
omegabytes: [d883ee3](https://github.com/Kong/koko/pull/364/commits/d883ee35fd047a928573720a02e7dc9999876fe8)

internal/metrics/v2/counter_test.go
omegabytes: Agree. I'm going to leave it since its a nit. I originally planned to test behavior when using multiple registries but I don't think that's necessary at the moment.

koko.yaml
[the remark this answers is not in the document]
omegabytes: [d15ec3d](https://github.com/Kong/koko/pull/364/commits/d15ec3d8f6a527c3bfd343cdec3690a8d55f1c4b)

internal/metrics/v2/histogram_test.go
omegabytes: [152f5a8](https://github.com/Kong/koko/pull/364/commits/152f5a8e502fc416b30574a645dd7f618a7d5887)

internal/metrics/v2/gauge_test.go
omegabytes: [cc5d6d7](https://github.com/Kong/koko/pull/364/commits/cc5d6d7753178dfab3aaab59957ddd8a53c0650c)

koko.yaml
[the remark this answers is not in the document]
tjasko: Leaving this conversation open for Harry in case he wants to respond. 🙂

internal/metrics/v2/gauge.go
tjasko: It's technically still valid, but yeah, it's not ideal. I don't see this being corrected, the indention is still very screwy & hard to tell the args being passed into `NewGaugeVec()` due to the formatting.

internal/server/kong/ws/config_service.go
tjasko: Saw that these were removed, do we want to just implement it in another PR then?

internal/metrics/v2/counter_test.go
tjasko: Yeah I don't think it's necessary anyway. Either way works with me.

internal/metrics/v2/counter_test.go
tjasko: Looks good!

You can use the defined metric type constants that the Prometheus protobuf spec defines btw:
```
require.Equal(t, io_prometheus_client.MetricType_COUNTER, family[0].Type)
```

internal/server/kong/ws/config_service.go
omegabytes: Sorry, I misunderstood "another commit" to mean "another PR". Let's go with another PR either way since these seem to replace the existing metrics, but the PR when handed off to me was duplicating rather than replacing. I'll handle the replacement in isolation.

internal/metrics/v2/counter_test.go
omegabytes: I can change it to this if you'd like
```
require.Equal(t, io_prometheus_client.MetricType_COUNTER.String(), family[0].Type.String())
```

internal/server/kong/ws/config_service.go
omegabytes: I'm working on a follow-up PR that implements various dp <-> cp metrics, I'll add this removed commit into those batch of changes. Branch here: https://github.com/Kong/koko/compare/feat/metrics-v2...feat/dp-metrics

tjasko: Work with me, that sounds reasonable anyway.

internal/metrics/v2/counter_test.go
tjasko: It's a bit of a nit, but the [example code](https://github.com/Kong/koko/pull/364/files#r1001067920) I posted above would be the ideal way to do it :)

omegabytes: [659dcb4](https://github.com/Kong/koko/pull/364/commits/659dcb4efbb0eeb7d300bd58b22b1ffb2b86b782)

internal/config/types.go
```diff
@@ -21,8 +21,7 @@ type ControlServer struct {
 
 // Metrics config.
 type Metrics struct {
-	// ClientType metrics client type e.g. prometheus, datadog.
-	ClientType string `yaml:"client_type" json:"client_type" env:"CLIENT_TYPE" env-default:"noop"`
```
tjasko, line 25: What's the plan for rolling this out, as this would remove the Datadog capabilities?

tjasko (commented):
Changes look good other than one comment that was not taken into account. 😀

internal/config/types.go
omegabytes: I can revert Harry's changes where it defaults to prometheus client (and default to DD), which would look like this:
```
	err := metrics.InitMetricsClient(logger.With(zap.String("component", "metrics-collector")), "datadog")
	if err != nil {
		return fmt.Errorf("init metrics client failure: %w", err)
	}

	defer metrics.Close()
	if config.Metrics.Enabled {
		metricsLogger := logger.With(zap.String("component", "metrics"))
		metricsHandler, err := metrics.CreateHandler(metricsLogger)
		if err != nil {
			return fmt.Errorf("create metrics handler failure: %w", err)
		}
		s, err := server.NewHTTP(server.HTTPOpts{
			Address: ":9090",
			Logger:  metricsLogger,
			Handler: serverUtil.HandlerWithRecovery(metricsHandler, metricsLogger),
		})
		if err != nil {
			return err
		}
		g.AddWithCtxE(s.Run)
	}
```
Or, I can do whatever work is necessary to ingest the prometheus emitted metrics into datadog.

omegabytes: [9264fa9](https://github.com/Kong/koko/pull/364/commits/9264fa95b62aa21fc12bae40048520bb79d4ecc5)

tjasko: How are we expecting this to be used if there's no way to configure it to use the newer metrics implementation? We should be feature-flagging this off.

omegabytes: @tjasko [e51578f](https://github.com/Kong/koko/pull/364/commits/e51578f5b513005597bc917b3ac20de8076c9f22)

omegabytes: Validated server is working locally 
```
/GET localhost:9090/metrics

# HELP go_gc_duration_seconds A summary of the pause duration of garbage collection cycles.
# TYPE go_gc_duration_seconds summary
go_gc_duration_seconds{quantile="0"} 8.416e-06
go_gc_duration_seconds{quantile="0.25"} 3.9667e-05
go_gc_duration_seconds{quantile="0.5"} 5.9041e-05
go_gc_duration_seconds{quantile="0.75"} 9.8332e-05
go_gc_duration_seconds{quantile="1"} 0.000186959
go_gc_duration_seconds_sum 0.000561706
go_gc_duration_seconds_count 8
...
```

internal/config/types.go
```diff
@@ -52,12 +58,14 @@ type TLS struct {
...
-	Control                 ControlServer `yaml:"control_server" json:"control_server" env-prefix:"KOKO_CONTROL_SERVER_"`
-	Database                Database      `yaml:"database" json:"database" env-prefix:"KOKO_DATABASE_"`
-	Metrics                 Metrics       `yaml:"metrics" json:"metrics" env-prefix:"KOKO_METRICS_"`
-	DisableAnonymousReports bool          `yaml:"disable_anonymous_reports" json:"disable_anonymous_reports" env-prefix:"KOKO_DISABLE_ANONYMOUS_REPORTS"` //nolint:lll
+	Log      Log           `yaml:"log" json:"log" env-prefix:"KOKO_LOG_"`
+	Admin    AdminServer   `yaml:"admin_server" json:"admin_server" env-prefix:"KOKO_ADMIN_SERVER_"`
+	Control  ControlServer `yaml:"control_server" json:"control_server" env-prefix:"KOKO_CONTROL_SERVER_"`
+	Database Database      `yaml:"database" json:"database" env-prefix:"KOKO_DATABASE_"`
+	// Deprecated
+	Metrics                 Metrics           `yaml:"metrics" json:"metrics" env-prefix:"KOKO_METRICS_"`
```
tjasko, line 66: Can we actually keep this around, deprecate the fields inside that are unused, and more Prometheus in there? I say that because there may be a global-scoped metric config that may impact all integrations in the future, and that'll at least set us up for that (just in case).

The struct would look something like:
```go
type Metrics struct {
    // Deprecated: This is kept around to enable the Datadog metrics integration
    // to continue to work, until we're ready to switch over to Prometheus.
	ClientType string            `yaml:"client_type" json:"client_type" env:"CLIENT_TYPE" env-default:"noop"`
    
    Prometheus PrometheusMetrics `yaml:"log" json:"log" env-prefix:"PROMETHEUS_"`
}

// PrometheusMetrics defines the configuration for setting up the Prometheus
// metrics exporter, which exposes an HTTP server returning metrics in the
// Prometheus exposition format under the `/metrics` HTTP path. 
type PrometheusMetrics struct {
	Enable  bool   `yaml:"enable" json:"enable" env:"ENABLE"`
	Address string `yaml:"address" json:"address" env:"ADDRESS" env-default:":9090"`
}
```

internal/config/types.go
```diff
@@ -20,11 +20,17 @@ type ControlServer struct {
...
+// Deprecated.
 type Metrics struct {
-	Enabled    bool   `yaml:"enabled" json:"enabled" env:"ENABLED"`
 	ClientType string `yaml:"client_type" json:"client_type" env:"CLIENT_TYPE" env-default:"noop"`
 }
 
+// PrometheusMetrics config.
+type PrometheusMetrics struct {
+	Enabled bool   `yaml:"enabled" json:"enabled" env:"ENABLED"`
+	Port    string `yaml:"port" json:"port" env:"PORT" env-default:":9090"`
```
tjasko, line 31: Thanks for doing this – That's cool with me, however, let's rename `Port` to `Address`, as it's a listen address, not just port.

internal/config/types.go
```diff
@@ -20,11 +20,17 @@ type ControlServer struct {
...
 // Metrics config.
+// Deprecated.
 type Metrics struct {
-	Enabled    bool   `yaml:"enabled" json:"enabled" env:"ENABLED"`
 	ClientType string `yaml:"client_type" json:"client_type" env:"CLIENT_TYPE" env-default:"noop"`
 }
 
+// PrometheusMetrics config.
+type PrometheusMetrics struct {
+	Enabled bool   `yaml:"enabled" json:"enabled" env:"ENABLED"`
```
tjasko, line 30: Let's actually rename this to just `Enable` to match that of the Postgres TLS config & the global TLS config... least we can get it to be consistent.

```suggestion
	Enable bool   `yaml:"enable" json:"enable" env:"ENABLE"`
```

omegabytes:
@tjasko [2420909](https://github.com/Kong/koko/pull/364/commits/24209094a77e2859e32db82e69b3f854e66b709b)

internal/cmd/run.go
```diff
@@ -63,29 +64,32 @@ func Run(ctx context.Context, config ServerConfig) error {
...
+			Handler: serverUtil.HandlerWithRecovery(serverUtil.HandlerWithLogger(metricsHandler, metricsLogger), metricsLogger),
 		})
 		if err != nil {
 			return err
 		}
 		g.AddWithCtxE(s.Run)
 	}
 
+	// TODO: Instrumentation with metrics client is deprecated and will be replaced by metricsv2 package.
+	if config.Metrics.ClientType == metrics.Datadog.String() {
```
tjasko, line 84: We'll need to ignore this from the linter:
```suggestion
	if config.Metrics.ClientType == metrics.Datadog.String() { //nolint:staticcheck
```

internal/config/testdata/good-0.yaml
```diff
@@ -22,8 +22,8 @@ database:
       hostname: "read-localhost"
 metrics:
   client_type: "noop"
-prometheus_metrics:
-  enabled: false
+  prometheus:
+    enabled: false
```
tjasko, line 26: We'll want to rename this to `enable`. This works because the default value is false anyway.

omegabytes: thanks for catching

tjasko (approved):
LGTM – this is a great step in the right direction, thanks for pushing this over the line @omegabytes

PR #444, opened 2022-09-22
refactor: add support to register additional FS for schemas loading

aboudreault:
This add the support to register additional FS for schemas loading. It also makes the`model.json.generator.registry` reusable instead of global.


internal/cmd/init_test.go
```diff
@@ -0,0 +1,7 @@
+package cmd
```
tjasko, line 1: Wouldn't this be better suited in `internal/test/util` or perhaps `internal/test`? I would also rename this file to `init.go` when moved there.

aboudreault: @tjasko so this means that in all files that require these schemas, I need to ensure to import that test/util module? am I right? (if not already used)

aboudreault: note that this is the main one: https://github.com/Kong/koko/pull/444/files#diff-7bc2adc8698fa2d6855fd6659a309ec9294ab00c02699e11b20f06862d46a689R184

this file intend is only to ensure that function is ran before running tests.

tjasko: Ah, we spoke about this on the side, but you can mind me. I wasn't fully thinking this through, and for whatever reason didn't notice some other changes being added.

tjasko (approved):
Changes LGTM, sorry for giving you the runaround here @aboudreault!

aboudreault:
@tjasko no worries, thanks all for the reviews and help.

PR #446, opened 2022-09-23
tests: e2e validation of plugin field updates for compatibility

omegabytes:
Recreates end-to-end tests used to validate updates to plugin fields that can occur when the version of a control plane and a data plane differ. These tests were removed in #368 due to a refactor.

### DIFF
--- a/internal/test/e2e/version_compatibility_test.go
+++ b/internal/test/e2e/version_compatibility_test.go
@@ -170,7 +170,7 @@ func TestVersionCompatibility(t *testing.T) {
 	})
 }
 
-func TestVersionCompatibility_FieldUpdatesEnsurePlugins(t *testing.T) {
+func TestVersionCompatibility_PluginFieldUpdates(t *testing.T) {
 	cleanup := run.Koko(t)
 	defer cleanup()
 
@@ -192,45 +192,42 @@ func TestVersionCompatibility_FieldUpdatesEnsurePlugins(t *testing.T) {
 
 	// Remove plugins that may not be expected due to data plane version or aren't subject to compatibility updates
 	expectedPluginsMap := make(map[string]VersionCompatibilityPlugins, 0)
-	for _, pluginTest := range VersionCompatibilityOSSPluginConfigurationTests {
-		addPlugin := true
-		if len(pluginTest.VersionRange) > 0 {
-			version := versioning.MustNewRange(pluginTest.VersionRange)
+	for _, plugin := range VersionCompatibilityOSSPluginConfigurationTests {
+		if plugin.FieldUpdateChecks == nil {
+			continue
+		}
+
+		if len(plugin.VersionRange) > 0 {
+			version := versioning.MustNewRange(plugin.VersionRange)
 			if !version(dataPlaneVersion) {
-				addPlugin = false
+				continue
 			}
 		}
-		if pluginTest.FieldUpdateChecks == nil {
-			addPlugin = false
-		}
-
-		if addPlugin {
-			expectedPluginsMap[pluginTest.Name] = pluginTest
-		}
+		expectedPluginsMap[plugin.Name] = plugin
 	}
 
-	// create plugin records in the CP + DP
-	for _, pluginTest := range expectedPluginsMap {
+	// create plugins
+	for _, plugin := range expectedPluginsMap {
 		var config structpb.Struct
-		if len(pluginTest.Config) > 0 {
-			require.Nil(t, json.ProtoJSONUnmarshal([]byte(pluginTest.Config), &config))
+		if len(plugin.Config) > 0 {
+			require.NoError(t, json.ProtoJSONUnmarshal([]byte(plugin.Config), &config))
 		}
 
 		p := &v1.Plugin{
 			Id:        uuid.NewString(),
-			Name:      pluginTest.Name,
+			Name:      plugin.Name,
 			Config:    &config,
 			Enabled:   wrapperspb.Bool(true),
 			Protocols: []string{"http", "https"},
 		}
 
 		pluginBytes, err := json.ProtoJSONMarshal(p)
-		require.Nil(t, err)
+		require.NoError(t, err)
 		res := admin.POST("/v1/plugins").WithBytes(pluginBytes).Expect()
 		res.Status(http.StatusCreated)
 	}
 
-	// fetch list of created plugin records from the DP
+	// fetch list of plugins from the DP
 	dataPlanePlugins := []*kongClient.Plugin{}
 	util.WaitFunc(t, func() error {
 		dataPlanePlugins, err = kongAdmin.Plugins.ListAll(ctx)
@@ -243,15 +240,15 @@ func TestVersionCompatibility_FieldUpdatesEnsurePlugins(t *testing.T) {
 		return err
 	})
 
+	require.Equal(t, len(expectedPluginsMap), len(dataPlanePlugins), "plugins configured count does not match")
 	dpPluginsMap := make(map[string]*kongClient.Plugin, 0)
 	for _, dpPlugin := range dataPlanePlugins {
 		dpPluginsMap[*dpPlugin.Name] = dpPlugin
 	}
-	require.Equal(t, len(expectedPluginsMap), len(dpPluginsMap), "plugins configured count does not match")
 
 	for name, expectedPlugin := range expectedPluginsMap {
 		t.Run(fmt.Sprintf("%s successfully updates fields", name), func(t *testing.T) {
-			require.Contains(t, dpPluginsMap, name, "failed to discover plugin")
+			require.Contains(t, dpPluginsMap, name, "plugin not present in the DP")
 			for rawVersion, updates := range expectedPlugin.FieldUpdateChecks {
 				parsedVersion := versioning.MustNewRange(rawVersion)
 				if parsedVersion(dataPlaneVersion) {
--- a/internal/test/e2e/version_compatibility_test.go
+++ b/internal/test/e2e/version_compatibility_test.go
@@ -7,7 +7,6 @@ import (
 	"errors"
 	"fmt"
 	"net/http"
-	"reflect"
 	"testing"
 
 	"github.com/gavv/httpexpect/v2"
@@ -20,6 +19,7 @@ import (
 	"github.com/kong/koko/internal/test/util"
 	"github.com/kong/koko/internal/versioning"
 	"github.com/stretchr/testify/require"
+	"github.com/tidwall/gjson"
 	"google.golang.org/protobuf/types/known/structpb"
 	"google.golang.org/protobuf/types/known/wrapperspb"
 )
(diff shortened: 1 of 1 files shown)


internal/test/e2e/version_compatibility_test.go
```diff
@@ -169,6 +170,102 @@ func TestVersionCompatibility(t *testing.T) {
...
+	require.NoError(t, err, "fetching Kong Gateway info")
+
+	dataPlaneVersion, err := versioning.NewVersion(kongClient.VersionFromInfo(info))
+	require.NoError(t, err)
+
+	admin := httpexpect.New(t, "http://localhost:3000")
+
+	// Remove plugins that may not be expected due to data plane version or aren't subject to compatibility updates
+	expectedPluginsMap := make(map[string]VersionCompatibilityPlugins, 0)
+	for _, pluginTest := range VersionCompatibilityOSSPluginConfigurationTests {
```
GGabriele, line 195: I think this can be rewritten in a more efficient way:

```
    for _, pluginTest := range VersionCompatibilityOSSPluginConfigurationTests {
		if pluginTest.FieldUpdateChecks == nil {
			continue
		}
		if len(pluginTest.VersionRange) > 0 {
			version := versioning.MustNewRange(pluginTest.VersionRange)
			if !version(dataPlaneVersion) {
				continue
			}
		}
		expectedPluginsMap[pluginTest.Name] = pluginTest
	}
```

This way you don't need the extra `addPlugin` and avoid cycles when not needed

internal/test/e2e/version_compatibility_test.go
```diff
@@ -169,6 +170,102 @@ func TestVersionCompatibility(t *testing.T) {
...
+		if pluginTest.FieldUpdateChecks == nil {
+			addPlugin = false
+		}
+
+		if addPlugin {
+			expectedPluginsMap[pluginTest.Name] = pluginTest
+		}
+	}
+
+	// create plugin records in the CP + DP
```
GGabriele, line 212:
```suggestion
	// create plugins
```

internal/test/e2e/version_compatibility_test.go
```diff
@@ -169,6 +170,102 @@ func TestVersionCompatibility(t *testing.T) {
...
+	for _, pluginTest := range expectedPluginsMap {
+		var config structpb.Struct
+		if len(pluginTest.Config) > 0 {
+			require.Nil(t, json.ProtoJSONUnmarshal([]byte(pluginTest.Config), &config))
```
GGabriele, line 216:
```suggestion
			require.NoError(t, json.ProtoJSONUnmarshal([]byte(pluginTest.Config), &config))
```

internal/test/e2e/version_compatibility_test.go
```diff
@@ -169,6 +170,102 @@ func TestVersionCompatibility(t *testing.T) {
...
+		p := &v1.Plugin{
+			Id:        uuid.NewString(),
+			Name:      pluginTest.Name,
+			Config:    &config,
+			Enabled:   wrapperspb.Bool(true),
+			Protocols: []string{"http", "https"},
+		}
+
+		pluginBytes, err := json.ProtoJSONMarshal(p)
+		require.Nil(t, err)
```
GGabriele, line 228:
```suggestion
		require.NoError(t, err)
```

internal/test/e2e/version_compatibility_test.go
```diff
@@ -169,6 +170,102 @@ func TestVersionCompatibility(t *testing.T) {
...
+		res := admin.POST("/v1/plugins").WithBytes(pluginBytes).Expect()
+		res.Status(http.StatusCreated)
+	}
+
+	// fetch list of created plugin records from the DP
```
GGabriele, line 233:
```suggestion
	// fetch list of plugins from the DP
```

internal/test/e2e/version_compatibility_test.go
```diff
@@ -169,6 +170,102 @@ func TestVersionCompatibility(t *testing.T) {
...
+	dataPlanePlugins := []*kongClient.Plugin{}
+	util.WaitFunc(t, func() error {
+		dataPlanePlugins, err = kongAdmin.Plugins.ListAll(ctx)
+		if err != nil {
+			t.Log("fetch plugin failed", err)
+		}
+		if len(dataPlanePlugins) == 0 {
```
GGabriele, line 240: You can replace this with the other check you have at line 250:

```
require.Equal(t, len(expectedPluginsMap), len(dataPlanePlugins), "plugins configured count does not match")
```

internal/test/e2e/version_compatibility_test.go
```diff
@@ -169,6 +170,102 @@ func TestVersionCompatibility(t *testing.T) {
...
+			return errors.New("plugins len was 0")
+		}
+		return err
+	})
+
+	dpPluginsMap := make(map[string]*kongClient.Plugin, 0)
+	for _, dpPlugin := range dataPlanePlugins {
+		dpPluginsMap[*dpPlugin.Name] = dpPlugin
+	}
+	require.Equal(t, len(expectedPluginsMap), len(dpPluginsMap), "plugins configured count does not match")
```
GGabriele, line 250: This can be removed if you move it up in the code, as per my other comment

internal/test/e2e/version_compatibility_test.go
```diff
@@ -169,6 +170,102 @@ func TestVersionCompatibility(t *testing.T) {
...
+
+	for name, expectedPlugin := range expectedPluginsMap {
+		t.Run(fmt.Sprintf("%s successfully updates fields", name), func(t *testing.T) {
+			require.Contains(t, dpPluginsMap, name, "failed to discover plugin")
```
GGabriele, line 254:
```suggestion
			require.Contains(t, dpPluginsMap, name, "plugin not present in the DP")
```

internal/test/e2e/version_compatibility_test.go
```diff
@@ -169,6 +170,102 @@ func TestVersionCompatibility(t *testing.T) {
...
+			for rawVersion, updates := range expectedPlugin.FieldUpdateChecks {
+				parsedVersion := versioning.MustNewRange(rawVersion)
+				if parsedVersion(dataPlaneVersion) {
+					for _, update := range updates {
+						want := fmt.Sprintf("%v", update.Value)
```
GGabriele, line 259: I think these can simply be (please double-check it!):

```
						want := update.Value
						have := dpPluginsMap[name].Config[update.Field]
```

omegabytes: I fumbled around with this for a bit before landing on the presented solution. Here's what happens if I don't stringify and compare:
```
Error:  Not equal: 
            expected: []string([]string{"kong.log.err('Hello Koko!')"})
            actual  : []interface {}([]interface {}{"kong.log.err('Hello Koko!')"})
```

omegabytes: I got some feedback from @tjasko as well. Here's the two paths I can take:
```
for _, update := range updates {
	want := update.Value
	have := dpPluginsMap[name].Config[update.Field]
	switch update.Value.(type) {
	case string:
		require.Equal(t, want, have, "failed to validate plugin updates")
	case []string:
		require.ElementsMatch(t, want, have, "failed to validate plugin updates")
	default:
		require.Fail(t, fmt.Sprintf("Unexpected type(s): want: %s, have: %s, ",
		    reflect.TypeOf(want), reflect.TypeOf(have)))
	}
}
```
Or the much less verbose:
```
for _, update := range updates {
	want := fmt.Sprintf("%v", update.Value)
	have := fmt.Sprintf("%v", dpPluginsMap[name].Config[update.Field])
	require.Equal(t, want, have, "failed to validate plugin updates")
}	
```
I'd rather go with the stringified version since these version compatibility handlers have so many conditional statements already. I'm making sure that I'm not introducing something flaky by taking a shortcut.

GGabriele: I believe we can happily accept verbosity for correctness. If we are absolutely sure that stringify-all-the-things works well in all cases, including the `[]string` one, then proceed. Otherwise, let's just write some more lines of code. You can always wrap the verification logic in an util function to avoid more nesting:

```
for _, update := range updates {
	want := update.Value
	have := dpPluginsMap[name].Config[update.Field]
	verifyUpdates(want, have) // or something like `require.NoError(t, verifyUpdates(want, have))`
}
```

omegabytes: >  can't we simply convert the values (expected and response) to JSON and compare that way?

I considered this, but this is also converting to a string so I would (lazily) just skip the marshaling steps.

I pushed an update that checks the type instead.

internal/test/e2e/version_compatibility_test.go
```diff
@@ -169,6 +171,107 @@ func TestVersionCompatibility(t *testing.T) {
...
+	})
+
+	require.Equal(t, len(expectedPluginsMap), len(dataPlanePlugins), "plugins configured count does not match")
+	dpPluginsMap := make(map[string]*kongClient.Plugin, 0)
+	for _, dpPlugin := range dataPlanePlugins {
+		dpPluginsMap[*dpPlugin.Name] = dpPlugin
+	}
+
+	for name, expectedPlugin := range expectedPluginsMap {
+		t.Run(fmt.Sprintf("%s successfully updates fields", name), func(t *testing.T) {
```
omegabytes, line 251: I added this test runner to more verbosely describe which plugin(s) fail and why. This includes the expected and actual config diffs. Previously, we were only returning a list of plugin names that failed.

internal/test/e2e/version_compatibility_test.go
```diff
@@ -169,6 +171,107 @@ func TestVersionCompatibility(t *testing.T) {
...
+			require.Contains(t, dpPluginsMap, name, "plugin not present in the DP")
+			for rawVersion, updates := range expectedPlugin.FieldUpdateChecks {
+				parsedVersion := versioning.MustNewRange(rawVersion)
+				if parsedVersion(dataPlaneVersion) {
+					for _, update := range updates {
+						want := update.Value
+						have := dpPluginsMap[name].Config[update.Field]
```
omegabytes, line 258: I just noticed this wont work when the Field value uses a path.to.field format. Eg: `update.Field = "response.body.time"`. I'll have to rewrite this using gjson or something...

internal/test/e2e/version_compatibility_test.go
mikefero: Your call, but this is going to be a maintenance issue in the future as we will need to keep adding to the case statement for each new type added.

Leaving final decision for @GGabriele.

GGabriele: Yeah I agree, I think you can easily re-use part of [the old logic](https://github.com/Kong/koko/pull/368/files#diff-5351ee7e6ddda0e9a38eee6852f369ac1c1fd272bba60c61f6ced419486215c3L456)...something like (definitely needs to be refactored):

```
                         for rawVersion, updates := range expectedPlugin.FieldUpdateChecks {
				parsedVersion := versioning.MustNewRange(rawVersion)
				config, err := json.ProtoJSONMarshal(dpPluginsMap[name])
				require.NoError(t, err, fmt.Sprintf("marshal %s plugin config: %v", name, err))
				configStr := string(config)
				if parsedVersion(dataPlaneVersion) {
					for _, update := range updates {
						res := gjson.Get(configStr, update.Field)
						if !res.Exists() || res.Value() != update.Value {
							return fmt.Errorf("expected %s, but got %s", update.Value, res.Value())
						}
					}
				}
			}
```

mikefero (approved):
Looks good to me, but I would still prefer not to validate differently based on types. Please get another +1 from @GGabriele.

GGabriele (approved):

tjasko (approved):

internal/test/e2e/version_compatibility_test.go
mikefero: You should be able to use [require.JSONEq](https://pkg.go.dev/github.com/stretchr/testify/require#JSONEq) to compare the JSON component instead of the objects themselves.

omegabytes: Thanks for the feedback. Looks like the solution with require.JSONEq will require creating of test plugin config fixtures to compare against. I'm going to stick with the approved solution, and we can revisit this if adding checks on types becomes unwieldy to maintain 😄

PR #449, opened 2022-09-26
feature: add referenceable fields

EJKinger:
This adds "referenceable" fields to the Certificate object.  This indicates that the field can include a secret via the [Kong secret reference format](https://docs.konghq.com/gateway/3.0.x/kong-enterprise/secrets-management/reference-format/).

https://github.com/Kong/kong/blob/release/3.0.x/kong/db/schema/entities/certificates.lua#L18-L21


internal/resource/certificate.go
```diff
@@ -53,68 +53,88 @@ func (r Certificate) Validate(ctx context.Context) error {
...
-			Errs: []*v1.ErrorDetail{
-				{
-					Type: v1.ErrorType_ERROR_TYPE_ENTITY,
-					Messages: []string{fmt.Sprintf("certificate and alternative certificate need to have "+
-						"different type (e.g. RSA and ECDSA), the provided "+
-						"certificates were both of the same type '%s'", cert.PublicKeyAlgorithm.String())},
+	var altCert *x509.Certificate
+	var altPubKey []byte
+
+	if !isReference(altCert) {
```
GGabriele, line 96:
```suggestion
	if !isReference(r.Certificate.CertAlt) {
```

internal/resource/helper.go
```diff
@@ -56,6 +57,17 @@ func parseURL(s *v1.Service) error {
 	return nil
 }
 
+func isReference(v interface{}) bool {
```
GGabriele, line 60: One you fix [this](https://github.com/Kong/koko/pull/449/files#r983158470), you can also simplify this function.

```
func isReference(v string) bool {
	r := regexp.MustCompile(typedefs.ReferencePattern)
	return r.MatchString(v)
}
```

internal/model/json/extension/config.go
```diff
@@ -34,6 +34,8 @@ type Config struct {
 	//
 	// This should be left empty in the event the resource is not exposed in the REST API.
 	ResourceAPIPath string `json:"resourceAPIPath,omitempty"`
+
+	Referenceable bool `json:"referenceable,omitempty"`
```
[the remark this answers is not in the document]
EJKinger: Added

internal/grpc/proto/kong/admin/model/v1/certificate.proto
```diff
@@ -3,31 +3,44 @@ syntax = "proto3";
 package kong.admin.model.v1;
 
 import "google/api/field_behavior.proto";
+import "google/protobuf/descriptor.proto";
 import "protoc-gen-openapiv2/options/annotations.proto";
 
 option go_package = "github.com/kong/koko/internal/gen/grpc/kong/admin/model/v1;v1";
 
+extend google.protobuf.FieldOptions {
+  bool referenceable = 50000;
```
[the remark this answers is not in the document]
EJKinger: Added a comment here.  As far as I can tell, we aren't doing any protobuf validations (lmk if I missed something), so the approach that I went with was to update the jsonschema requirements as well as the Certificate struct validations.

internal/resource/certificate.go
EJKinger: done

internal/resource/helper.go
EJKinger: done

internal/grpc/proto/kong/admin/model/v1/certificate.proto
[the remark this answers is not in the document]
tjasko: That's correct, we're not using anything like [protoc-gen-validate](https://github.com/envoyproxy/protoc-gen-validate) today, and well, wouldn't exactly make sense unless we can codegen it to JSON schema.

These Protobuf field options can be used programmatically, but if we're not planning on that, should we even bother adding it to the Protobuf field definition in the first place?

internal/resource/helper.go
```diff
@@ -56,6 +57,12 @@ func parseURL(s *v1.Service) error {
 	return nil
 }
 
+func isReference(v string) bool {
+	// https://github.com/Kong/kong/blob/5d721ac9ae1df36049013599d0253bd39cb6f758/kong/pdk/vault.lua#L320-L328
+	r := regexp.MustCompile(typedefs.ReferencePattern)
```
tjasko, line 62: Can we move this outside the function as an un-exported variable, to not compile the same regex every time? 

The other advantage is if `typedefs.ReferencePattern()` is an invalid regex, it'll panic almost immediately during application init instead of at function runtime (the `Must*()` functions generally are reserved for this exact use-case, as otherwise, we wouldn't want any possibility of panics).

internal/grpc/proto/kong/admin/model/v1/certificate.proto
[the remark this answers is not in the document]
EJKinger: The main reason I included it was so that any clients that use our proto definitions will have something to work with programatically.

internal/resource/helper.go
EJKinger: good catch, done.

GGabriele (approved):

tjasko (approved):

internal/grpc/proto/kong/admin/model/v1/certificate.proto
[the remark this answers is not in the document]
tjasko: Reasonable enough with me!

PR #453, opened 2022-09-27
Revert "set multiplatform dev build"

javierguerragiraldez:
Reverts Kong/koko#447

turns out it's unsupported until changed driver


tjasko:
@javierguerragiraldez We can get this closed out once we have confirmed the build is back to working as expected.

javierguerragiraldez:
the issue was fixed forward in #454

PR #450, opened 2022-09-27
refactor: various improvements to the persistence store integration

tjasko:
This is a refactor of the persistence store/DB integration, to reduce some spaghetti code, and to assist when creating new persistence store integrations, like with the recent MySQL integration work going on.

This change also allows for using the `KOKO_DATABASE_*` environment variables during E2E tests, to allow for more configurability when running locally. The legacy `KOKO_TEST_DB` environment variable is still supported, but I imagine we'll want to deprecate it. To reduce scope of these changes, I did not want to do that in this PR.

PR #455, opened 2022-09-27
feat: MySQL DB support

tjasko:
This change introduces support for MySQL >= 8.0.17.

—

**Work remaining/completed:**
- [x] Throw up PR to handle some refactoring of the persistence store integration.
  - See #450.
- [x] Implement MySQL persistence store support.
  - [x] Support tag-based listing (ensuring indexes are hit).
- [x] Run MySQL part of the CI test suite.
  - See #458.
- [x] Test in AWS RDS to ensure compatibility.
  - I tested MySQL with their minimum supported version of 8.0.23 and the latest supported version of 8.0.30. All works as expected.
- [x] Complete & verify TLS connection support.

### DIFF
--- a/internal/config/db.go
+++ b/internal/config/db.go
@@ -5,6 +5,7 @@ import (
 	"time"
 
 	"github.com/kong/koko/internal/db"
+	"github.com/kong/koko/internal/persistence/mysql"
 	"github.com/kong/koko/internal/persistence/postgres"
 	"github.com/kong/koko/internal/persistence/sqlite"
 	"go.uber.org/zap"
@@ -18,8 +19,29 @@ type Database struct {
 
 	QueryTimeout string `yaml:"query_timeout" json:"query_timeout" env:"QUERY_TIMEOUT" env-default:"5s"`
 
+	MySQL    MySQL    `yaml:"mysql" json:"mysql" env-prefix:"MYSQL_"`
 	SQLite   SQLite   `yaml:"sqlite" json:"sqlite" env-prefix:"SQLITE_"`
 	Postgres Postgres `yaml:"postgres" json:"postgres" env-prefix:"POSTGRES_"`
+}
+
+// MySQL defines configuration for using MySQL as the persistent store.
+type MySQL struct {
+	DBName      string           `yaml:"db_name" json:"db_name" env:"DB_NAME"`
+	Hostname    string           `yaml:"hostname" json:"hostname" env:"HOSTNAME"`
+	ReadReplica MySQLReadReplica `yaml:"read_replica" json:"read_replica" env-prefix:"READ_REPLICA_"`
+	Port        int              `yaml:"port" json:"port" env:"PORT"`
+	User        string           `yaml:"user" json:"user" env:"USER"`
+	Password    string           `yaml:"password" json:"password" env:"PASSWORD"`
+	TLS         TLS              `yaml:"tls" json:"tls" env-prefix:"TLS_"`
+
+	// See the `Params` field on mysql.Opts for more info.
+	Params map[string]string `yaml:"params" json:"params" env:"PARAMS"`
+}
+
+// MySQLReadReplica defines configuration for specifying a single read-replica.
+// This configuration overrides fields set in config.MySQL.
+type MySQLReadReplica struct {
+	Hostname string `yaml:"hostname" json:"hostname" env:"HOSTNAME"`
 }
 
 // Postgres defines configuration for using Postgres as the persistent store.
@@ -49,6 +71,32 @@ type PostgresReadReplica struct {
 type SQLite struct {
 	Filename string `yaml:"filename" json:"filename" env:"FILENAME"`
 	InMemory bool   `yaml:"in_memory" json:"in_memory" env:"IN_MEMORY"`
+}
+
+// Opts returns the options required to instantiate a MySQL persistence.Persister.
+func (c *MySQL) Opts() (mysql.Opts, error) {
+	if err := c.TLS.fetchFileContents(); err != nil {
+		return mysql.Opts{}, err
+	}
+
+	opts := mysql.Opts{
+		DBName:           c.DBName,
+		Hostname:         c.Hostname,
+		ReadOnlyHostname: c.ReadReplica.Hostname,
+		Params:           c.Params,
+		Port:             c.Port,
+		User:             c.User,
+		Password:         c.Password,
+
+		// TLS options.
+		EnableTLS:                c.TLS.Enable,
+		RootCA:                   c.TLS.RootCA,
+		Certificate:              c.TLS.Certificate,
+		Key:                      c.TLS.Key,
+		SkipHostnameVerification: c.TLS.SkipHostnameVerification,
+	}
+
+	return opts, nil
 }
 
 // Opts returns the options required to instantiate a Postgres persistence.Persister.
@@ -89,10 +137,16 @@ func ToDBConfig(config Database, logger *zap.Logger) (db.Config, error) {
 		return db.Config{}, fmt.Errorf("failed to parse DB query timeout: %w", err)
 	}
 
+	mysqlOpts, err := config.MySQL.Opts()
+	if err != nil {
+		return db.Config{}, fmt.Errorf("unable to set MySQL DB options: %w", err)
+	}
+
 	return db.Config{
 		Dialect:      config.Dialect,
 		QueryTimeout: queryTimeout,
 		Logger:       logger,
+		MySQL:        mysqlOpts,
 		Postgres:     config.Postgres.Opts(),
 		SQLite:       config.SQLite.Opts(),
 	}, nil
--- a/internal/config/types.go
+++ b/internal/config/types.go
@@ -1,5 +1,10 @@
 package config
 
+import (
+	"fmt"
+	"os"
+)
+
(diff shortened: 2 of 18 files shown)


internal/persistence/mysql/mysql.go
```diff
@@ -0,0 +1,304 @@
...
+	// NOTE: The `->` & `->>` operators shall never be used, as they are not supported in MariaDB.
+	// Instead, `JSON_EXTRACT()` should be used in its place.
+	const tagsSelector = "JSON_EXTRACT(value, '$.object.tags')"
+	if exprFunction == operators.LogicalOr {
+		query = query.Where("JSON_OVERLAPS("+tagsSelector+", ?)", tagsJSON)
+	} else {
+		query = query.Where("JSON_CONTAINS("+tagsSelector+", ?, '$')", tagsJSON)
+	}
+
+	return query, nil
```
tjasko, line 303: For the record, I did confirm this is hitting the index, using `EXPLAIN ANALYZE` in MySQL 8.0.18. Example of `JSON_OVERLAPS` hitting the index below:
```
-> Window aggregate with buffering  (actual time=0.391..0.391 rows=1 loops=1)
    -> Filter: ((store.`key` like 'c/default/o/node/%') and json_overlaps(cast(json_extract(store.`value`,_utf8mb4'$.object.tags') as char(128) array),json'["tag2"]'))  (cost=0.71 rows=1) (actual time=0.138..0.383 rows=1 loops=1)
        -> Index range scan on store using tags_idx  (cost=0.71 rows=1) (actual time=0.134..0.374 rows=31 loops=1)
```

internal/db/sql/mysql/0001_init.up.sql
```diff
@@ -0,0 +1,6 @@
+CREATE TABLE IF NOT EXISTS store (
+  `key` VARCHAR(512) NOT NULL,
```
tjasko, line 2: `key` is a reserved word in MySQL, so we need to escape it. I considered renaming it, but it's not that big of a deal. If anyone has any strong opinions, we can absolutely rename it to avoid the escaping.

internal/config/types.go
```diff
@@ -36,3 +59,36 @@ type Config struct {
...
+		{"certificate", t.CertificateFile, &t.Certificate},
+		{"key", t.KeyFile, &t.Key},
+	} {
+		// No filename set for the field, so we can ignore.
+		if d.filename == "" {
+			continue
+		}
+
+		if *d.ptr != "" {
+			return fmt.Errorf("cannot set both %s file & raw string", d.name)
```
GGabriele, line 82: Would it be useful to also print the filename here and in the next error?

internal/db/db.go
```diff
@@ -28,6 +34,11 @@ func NewSQLDBFromConfig(config Config) (*sql.DB, error) {
 	var err error
 
 	switch config.Dialect {
+	case DialectMariaDB:
```
GGabriele, line 37: Maybe I'm missing some context: why do we want to include `MariaDB` here if it's not supported yet?

GGabriele:
I guess you'd also want to update the matrix [here](https://github.com/Kong/koko/blob/main/docs/koko-vs-kong.md)

internal/db/db.go
tjasko: Wanted to explicitly have it in the code to say it's not supported vs. having it just be in the documentation. This way, if someone were to set the dialect to `mariadb`, it would error out.

internal/config/types.go
tjasko: For this error, maybe. The error gets wrapped into the "unable to set MySQL DB options" error, and I was thinking that's more than enough as it refers to the config. No opposition to adding the file name here though.

As for the next error, nah, the Golang open file tooling produces an error with the file path:
```
unable to set MySQL DB options: unable to read root CA file contents: open /this/file/path/does/not/exist.pem: no such file or directory
```

tjasko:
> @GGabriele I guess you'd also want to update the matrix [here](https://github.com/Kong/koko/blob/main/docs/koko-vs-kong.md)

Indeed, wasn't sure if we wanted to do that in this PR or do it afterward in case anything changed (like supported versions). At this point, I'm fine with adding it as things likely wouldn't be changing there, so will go ahead & update that documentation.

internal/db/sql/mysql/0001_init.up.sql
```diff
@@ -0,0 +1,6 @@
+CREATE TABLE IF NOT EXISTS store (
```
tjasko, line 1: - [ ] Add down migration

internal/config/db.go
```diff
@@ -18,10 +19,31 @@ type Database struct {
...
+	Password    string           `yaml:"password" json:"password" env:"PASSWORD"`
+	TLS         TLS              `yaml:"tls" json:"tls" env-prefix:"TLS_"`
+
+	// See the `Params` field on mysql.Opts for more info.
+	Params map[string]string `yaml:"params" json:"params" env:"PARAMS"`
+}
+
+// MySQLReadReplica defines configuration for specifying a single read-replica.
+// This configuration overrides fields set in config.MySQL.
+type MySQLReadReplica struct {
```
EJKinger, line 43: In addition to hostname, we should be able to configure distinct port, user, password, and tls for the read replica.  I'm not sure about Params, but I assume this as well.

internal/config/db.go
```diff
@@ -18,10 +19,31 @@ type Database struct {
...
+	MySQL    MySQL    `yaml:"mysql" json:"mysql" env-prefix:"MYSQL_"`
 	SQLite   SQLite   `yaml:"sqlite" json:"sqlite" env-prefix:"SQLITE_"`
 	Postgres Postgres `yaml:"postgres" json:"postgres" env-prefix:"POSTGRES_"`
 }
 
+// MySQL defines configuration for using MySQL as the persistent store.
+type MySQL struct {
+	DBName      string           `yaml:"db_name" json:"db_name" env:"DB_NAME"`
+	Hostname    string           `yaml:"hostname" json:"hostname" env:"HOSTNAME"`
+	ReadReplica MySQLReadReplica `yaml:"read_replica" json:"read_replica" env-prefix:"READ_REPLICA_"`
```
EJKinger, line 31: Does this make more sense as an array, since we may want to configure n number of read-replicas?

EJKinger:
/koko.yaml doesn't have an example configuration for `mysql`.

internal/config/db.go
tjasko: Harry has specifically disallowed arrays: https://github.com/Kong/koko/blob/main/internal/config/types_test.go#L11

Additionally, I'm just copying what was done with Postgres:
https://github.com/Kong/koko/blob/main/internal/config/db.go#L30

I'd say if we'd want to expand this, we'd need to punt that off to another issue.

internal/config/db.go
tjasko: It's documented that read replica shares config with the primary:
https://github.com/Kong/koko/blob/feat/mysql-support/internal/persistence/mysql/options.go#L39-L41

There's a lot of duplicate config due to the app DTO -> persistence store DTOs that we have, and I was trying to keep most things in one place.

Additionally, this is how Postgres works today as well, so I didn't want to diverge from that.

tjasko:
> @EJKinger /koko.yaml doesn't have an example configuration for MySQL.

Ah, nice catch. Will add one.

internal/db/persister.go
```diff
@@ -14,6 +16,11 @@ func NewPersister(config Config) (persistence.Persister, error) {
 		err       error
 	)
 	switch config.Dialect {
+	case DialectMariaDB:
+		// See mysql.MySQL on why MariaDB is not supported.
+		err = errors.New("MariaDB is currently unsupported")
```
GGabriele, line 21: Since we're using this string 4 times in this package, what do you think about creating an error and reuse it?

koko.yaml
```diff
@@ -2,8 +2,57 @@ log:
...
+        ...
+        -----END CERTIFICATE-----
+      key: |
+        -----BEGIN RSA PRIVATE KEY-----
+        ...
+        -----END RSA PRIVATE KEY-----
+
+    # Optional hostname to declare a single read-replica. Shares the
+    # same connectivity settings as the primary DB settings above.
+    read_replica: another-hostname
```
GGabriele, line 55:
```suggestion
    read_replica:
        hostname: another-hostname
```

internal/db/persister.go
tjasko: Yeah, hopefully, it won't last long, but went ahead and added it in [e00582d](https://github.com/Kong/koko/pull/455/commits/e00582d58f8f394a90387480c6631084ea8c450a).

koko.yaml
tjasko: Ah, oops, nice catch! Fixed in https://github.com/Kong/koko/pull/455/commits/1fea4b175307e48473d02add2ca958cf9de6d89a.

EJKinger (approved):

tjasko:
Thanks @EJKinger! Will hold the merge until @GGabriele is able to give it another look, to give him the ability to ensure his comments were addressed.

GGabriele (approved):
Great work here!

tjasko:
> @GGabriele Great work here!

Many thanks! 😊

PR #458, opened 2022-09-28
tests: add MySQL to the test suite in CI

tjasko:
This change utilizes the MySQL integration introduced in #455 & adds testing the MySQL integration to the test suite.


GGabriele (approved):
🚢 it

PR #461, opened 2022-10-03
Add atc-expression routes with syntax verification

javierguerragiraldez:


internal/resource/route.go
```diff
@@ -638,6 +684,22 @@ func init() {
 					},
 				},
 			},
+			{
+				Title: ExpressionRouteRuleTitle,
+				Description: "When 'expression' is defined, 'priority' is required and " +
+					"'snis', 'sources' or 'destinations' cannot be set.",
+				If: &generator.Schema{
+					Required: []string{"expression"},
+				},
```
hbagdi, line 693: Let's split this out into two different rules.
- priority is required
- all other fiends cannot be set.


Question: are host, paths, methods and headers allowed to be set when expression is used?

scripts/install-deps.sh
```diff
@@ -0,0 +1,4 @@
+#!/bin/bash -e
+
+curl -sSfL "https://raw.githubusercontent.com/Kong/go-atc-router/trunk/build-deps.sh" \
```
hbagdi, line 3: Merge blocker: use 'main' branch and no forks in the script first.
This script probably belongs in this repo itself perhaps?

internal/resource/route.go
```diff
@@ -638,6 +684,22 @@ func init() {
...
+				Then: &generator.Schema{
+					Required: []string{"priority"},
+					Properties: map[string]*generator.Schema{
+						"snis":         {Not: &generator.Schema{}},
+						"sources":      {Not: &generator.Schema{}},
+						"destinations": {Not: &generator.Schema{}},
```
GGabriele, line 699: Other than these, the following fields cannot be set with `expression`: `paths`, `hosts`, `headers`, `methods`, `regex_priority`, `path_handling`

javierguerragiraldez: done.  had to make the `r.ProcessDefaults()` method deal differently for expression routes.

javierguerragiraldez: also made the messages more specific, otherwise it was too long to be helpful.

internal/resource/route.go
javierguerragiraldez: done.

some of the other fields were hard to check because they were set by the ProcessDefaults method, so i updated that too.

internal/resource/route.go
GGabriele: Yeah that looks much better now, thanks!

internal/resource/route.go
```diff
@@ -102,17 +116,58 @@ func (r Route) ProcessDefaults(ctx context.Context) error {
...
 		}
 	}
 	defaultID(&r.Route.Id)
 	return nil
 }
 
+func buildSchema() *goatcrouter.Schema {
+	schema := goatcrouter.NewSchema()
+
+	for _, fieldname := range []string{
```
tjasko, line 140: nit: might be better to use `map[goatcrouter.FieldType][]string`

internal/resource/route.go
```diff
@@ -102,17 +116,58 @@ func (r Route) ProcessDefaults(ctx context.Context) error {
...
 func init() {
+	jsonschema.Formats["route-expression"] = func(v interface{}) bool {
+		expression, ok := v.(string)
+		if !ok {
+			return false
+		}
+
+		router := goatcrouter.NewRouter(cachedSchema)
+		defer router.Free()
+		err := router.AddMatcher(0, uuid.Must(uuid.NewV4()), expression)
```
tjasko, line 167: Why the random UUID?

scripts/install-deps.sh
tjasko, line 3: Have we thought about local development requirements? What would the process look like if we needed to make updates to `go-atc-router` & need to have multiple versions in the mix?

tjasko: Will need to update the `go-atc-router` Go module version once it's gone through the code review process too, as we'd want to be pointing to a specific tag too.

internal/resource/route.go
```diff
@@ -5,13 +5,16 @@ import (
 	"fmt"
 	"net/http"
 
+	"github.com/gofrs/uuid"
 	"github.com/imdario/mergo"
+	goatcrouter "github.com/kong/go-atc-router"
```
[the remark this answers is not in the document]
javierguerragiraldez: you're right, nobody wants crusty French goats.

internal/resource/route.go
javierguerragiraldez: agreed.  that was a direct translation of the current Lua code, and should be kept in sync with that.  shouldn't be difficult as long as the list of predefined fields is visible in any syntax highlighter.

internal/resource/route.go
javierguerragiraldez: the `AddMatcher()` method requires an UUID to handle expressions.  the Lua code uses the route entity id, but in this JsonSchema callback we don't have access to other fields.

We don't actually use the id because we just want to verify that the expression gets compiled without error and drop the whole router object immediately.  since each router only ever gets a single expression, we could use the same predefined id every time.  either by moving the uuid creation to the init phase, or by using a literal constant.

but if performance results a real concern, we should get the internal Rust `parse` and `ast.validate` functions exposed and thus avoid the creation of a router object.

tjasko: Got it, makes sense, just seemed a bit strange while reading the code. Appreciate the explanation. I'm assuming there's not much of a performance improvement using atc-router's own internal `parse()` & `ast.validate()` functions.

scripts/install-deps.sh
hbagdi: @javierguerragiraldez Any update?

.github/workflows/integration.yaml
```diff
@@ -32,6 +32,20 @@ jobs:
...
+        path: |
+          /tmp/libatc_router*
+        key: ${{ runner.os }}-go-atc-router-${{ hashFiles('**/go.sum') }}
+    - name: Install cached library
+      run: |
+        if [ -e /tmp/libatc_router* ]; then
+          sudo install /tmp/libatc_router* /usr/local/lib;
+        else
+          make install-deps
+        fi
```
hbagdi, line 48: This is smelly.
The CI environment should be responsible for restoring the cache if there exists one.
The script (make target) should be responsible for checking if the compiled library exists, if it is up to date, if not do its thing. The workflow on a developer's workstation and in CI should be identical.

internal/resource/route.go
```diff
@@ -6,13 +6,16 @@ import (
 	"net/http"
 	"strings"
 
+	"github.com/gofrs/uuid"
```
hbagdi, line 9: Merge blocker. Please use google/uuid instead in the upstream library and remove this dependency from this patch.

internal/resource/route.go
```diff
@@ -663,6 +728,53 @@ func init() {
...
+			},
+			{
+				Title: ExpressionRouteRuleTitle,
+				If: &generator.Schema{
+					Required: []string{"expression"},
+				},
+				Then: &generator.Schema{
+					Properties: map[string]*generator.Schema{
+						"hosts": {Not: &generator.Schema{
+							Description: "When 'expression' is defined, the field 'hosts' must not be set.",
```
hbagdi, line 749: Can we please generate this string programmatically to avoid the repetition?

scripts/install-deps.sh
javierguerragiraldez: i want to simplify a few things on `go-atc-`router`, especially those that move the library around.  part of the complexity is that the requirements for dynamic vs static linking are different (and both have ugly corners).

.github/workflows/integration.yaml
javierguerragiraldez: it was the simplest solution for dynamic linking, since the library must be "installed" in a well known place, but the standard cache action refuses to touch those directories.
in fact this is one of the big disadvantages of dynamic linking.  last time, there was some issues for fully static, but i think they can be solved a bit cleaner that this.

internal/resource/route.go
javierguerragiraldez: done

internal/resource/route.go
javierguerragiraldez: done

internal/resource/route.go
```diff
@@ -681,6 +758,28 @@ func init() {
...
+				Description: "When 'expression' is defined, 'priority' is required.",
+				If: &generator.Schema{
+					Required: []string{"expression"},
+				},
+				Then: &generator.Schema{
+					Required: []string{"priority"},
+				},
+			},
+			{
+				Title: ExpressionRouteRuleTitle,
```
GGabriele, line 772: Another rule should be added: `priority` can only be set with `expression` flavour

GGabriele:
Code looks good to me, except for one schema rule that we need to add for the `priority` field.

I couldn't test locally though, because of this error...would need to investigate:

```
# github.com/kong/koko/internal/cmd.test
/opt/homebrew/Cellar/go/1.19.2/libexec/pkg/tool/darwin_arm64/link: running clang failed: exit status 1
ld: unknown option: -Bstatic
clang: error: linker command failed with exit code 1 (use -v to see invocation)
```

Can you also rebase pls?

internal/resource/route.go
javierguerragiraldez: done

GGabriele (approved):

PR #462, opened 2022-10-06
chore: updating govulncheck to run with the latest Go 1.19 version

tjasko:
This matches our build logic, as we also currently build using the latest Go 1.19 version as well.

This is done to correct the following vulnerability:

```
Vulnerability #1: GO-2022-1039
  Programs which compile regular expressions from untrusted
  sources may be vulnerable to memory exhaustion or denial of
  service. The parsed regexp representation is linear in the size
  of the input, but in some cases the constant factor can be as
  high as 40,000, making relatively small regexps consume much
  larger amounts of memory. After fix, each regexp being parsed is
  limited to a 256 MB memory footprint. Regular expressions whose
  representation would use more space than that are rejected.
  Normal use of regular expressions is unaffected.

  Found in: regexp/syntax@go1.19.1
  Fixed in: regexp/syntax@go1.19.2
  More info: https://pkg.go.dev/vuln/GO-2022-1039
```

### DIFF
--- a/.github/workflows/security.yaml
+++ b/.github/workflows/security.yaml
@@ -7,7 +7,7 @@ jobs:
     - name: Set up Go
       uses: actions/setup-go@v3
       with:
-        go-version: '^1.19.1'
+        go-version: '^1.19'
     - name: Checkout repository
       uses: actions/checkout@v3
     - name: make install-tools
--- a/.github/workflows/security.yaml
+++ b/.github/workflows/security.yaml
@@ -8,6 +8,7 @@ jobs:
       uses: actions/setup-go@v3
       with:
         go-version: '^1.19'
+        check-latest: true
     - name: Checkout repository
       uses: actions/checkout@v3
     - name: make install-tools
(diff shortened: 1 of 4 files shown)


tjasko:
This will start passing once the below PR is merged:
https://github.com/actions/go-versions/pull/60

aboudreault (dismissed):

GGabriele (dismissed):
Thanks sir!

.github/workflows/security.yaml
```diff
@@ -7,7 +7,8 @@ jobs:
     - name: Set up Go
       uses: actions/setup-go@v3
       with:
-        go-version: '^1.19.1'
+        go-version: '^1.19'
+        check-latest: true
```
tjasko, line 11: Yes, using `check-latest` will cause it to make a call out to [go-versions](https://github.com/actions/go-versions) to see if it's the latest version, but that's a quick call and once it's in cache, it'll be fairly performant anyway:
https://github.com/actions/setup-go/blob/v3.3.0/src/installer.ts#L40

This will allow us to just re-trigger off the build in the event there's a security concern, instead of needing to find a way to bust the action's cache, alter the go-version, etc.

GGabriele (approved):

PR #464, opened 2022-10-08
feat: add vault support

EJKinger:
This PR adds a vault model and service for managing vaults.
Vaults include custom prefixes for managing the data plane's
access to environment variable based secrets.
This adds feature parity with kong running in control plane
mode.


internal/grpc/proto/kong/admin/model/v1/vault.proto
```diff
@@ -0,0 +1,21 @@
...
+
+message Vault {
+  string id = 1;
+  int32 created_at = 2;
+  int32 updated_at = 3;
+  string prefix = 4 [(google.api.field_behavior) = REQUIRED];
+  string name = 5 [(google.api.field_behavior) = REQUIRED];
+  string description = 6;
+  message Config {
+    string prefix = 1;
```
GGabriele, line 17: This only supports `env` vault, but i believe we should support all those available in the DP today, namely `aws`, `gcp` and `hcv`. Each of these have different required configuration and validation.

  https://docs.konghq.com/gateway/latest/kong-enterprise/secrets-management/backends/

internal/grpc/proto/kong/admin/model/v1/vault.proto
```diff
@@ -0,0 +1,21 @@
...
+  }
+  Config config = 7;
```
GGabriele, line 19:
```suggestion
  map<string, string> config = 7;
```

internal/resource/vault.go
```diff
@@ -0,0 +1,134 @@
...
+		Type: "object",
+		Properties: map[string]*generator.Schema{
+			"prefix": {Type: "string"},
+		},
+		AdditionalProperties: &falsy,
+	}
+	vaultSchema := &generator.Schema{
+		Properties: map[string]*generator.Schema{
+			"id":     typedefs.ID,
+			"prefix": {Type: "string"},
```
GGabriele, line 88: Can we add some pattern to match? https://github.com/Kong/kong/blob/master/kong/db/schema/entities/vaults.lua#L59

internal/grpc/proto/kong/admin/model/v1/vault.proto
EJKinger: This is done.

internal/resource/vault.go
EJKinger: This is done.

internal/resource/vault.go
```diff
@@ -0,0 +1,134 @@
...
+		return NewVault()
+	})
+	if err != nil {
+		panic(err)
+	}
+
+	envConfigSchema := &generator.Schema{
+		Type: "object",
+		Properties: map[string]*generator.Schema{
+			"prefix": {Type: "string"},
```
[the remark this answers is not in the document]
EJKinger: This is done.

internal/grpc/proto/kong/admin/model/v1/vault.proto
EJKinger: @GGabriele https://github.com/Kong/koko/pull/464/files#diff-85f2673ddcf702a04020c0c18ca5ee9eb1988643b86742d47a22c74d71d5fdf4R81-R85 to ensure that configuration remains compatible.

internal/resource/vault_test.go
```diff
@@ -0,0 +1,167 @@
...
+				v.Vault.Prefix = "UNKNOWN_"
+				return v
+			},
+			wantErr: true,
+			Errs: []*model.ErrorDetail{
+				{
+					Type:  model.ErrorType_ERROR_TYPE_FIELD,
+					Field: "name",
+					Messages: []string{
+						"value must be \"env\"",
```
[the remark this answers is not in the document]
EJKinger: This is added.

internal/resource/vault.go
```diff
@@ -0,0 +1,158 @@
...
+					},
+					Description: fmt.Sprintf("must not be any of %v or %v", vaultTypes, enterpriseVaultTypes),
+				},
+			},
+			"name": {
+				Type: "string",
+				Enum: vaultTypes,
+			},
+			"config": {
+				Type: "object",
```
GGabriele, line 120: Since this object is going to be defined by some rule based on the vault name, this should probably be written like this:

```
			"config": {
				Type:                 "object",
				AdditionalProperties: &truthy,
			},
```

internal/server/kong/ws/config/compat/compatibility_table.go
```diff
@@ -834,6 +834,20 @@ var (
 				},
 			},
 		},
+		{
+			Metadata: config.ChangeMetadata{
+				ID:          config.ChangeID("P135"),
+				Severity:    config.ChangeSeverityError,
+				Description: standardCoreEntityFieldsMessage("vault", []string{"*"}, versions300AndAbove),
```
GGabriele, line 841: The [standardCoreEntityFieldsMessage](https://github.com/Kong/koko/blob/main/internal/server/kong/ws/config/compat/compatibility_table.go#L46-L55) looks more related to entities fields, but the message for Vaults should be about the whole entity. Something like (please rephrase in good inglish :D ):

```
func standardCoreEntityMessage(entityName string, versionWithFeatureSupport string) string {
	return fmt.Sprintf("The '%s' entity is being used, "+
		"but Kong Gateway versions < %s do not support this entity. "+
		entityName,
		versionWithFeatureSupport,
	)
}
```

WDYT?

internal/grpc/proto/kong/admin/service/v1/vault.proto
```diff
@@ -0,0 +1,85 @@
...
+  model.v1.Vault item = 1;
+}
+
+message UpsertVaultRequest {
+  model.v1.Vault item = 1;
+  model.v1.RequestCluster cluster = 2;
+}
+
+message UpsertVaultResponse {
+  model.v1.Vault item = 1;
```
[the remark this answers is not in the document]
EJKinger: It seems like generally with other resources, when the request includes a cluster, the response doesn't include the details about the cluster in the request.

EJKinger:
Add e2e test to ensure that a vault entity can be successfully used by the dp.

internal/server/kong/ws/config/compat/compatibility_table.go
EJKinger: This is done

internal/server/kong/ws/config/compat/compatibility_table.go
```diff
@@ -834,6 +842,20 @@ var (
 				},
 			},
 		},
+		{
+			Metadata: config.ChangeMetadata{
+				ID:          config.ChangeID("P135"),
+				Severity:    config.ChangeSeverityError,
+				Description: standardCoreEntityMessage("vault", "3.0.0"),
```
GGabriele, line 849:
```suggestion
				Description: standardCoreEntityMessage("vault", "3.0"),
```

It seems we are consistently using `3.0` for update messages in this file.

EJKinger: Updated this.

internal/server/kong/ws/config/compat/compatibility_table.go
```diff
@@ -834,6 +842,20 @@ var (
 				},
 			},
 		},
+		{
+			Metadata: config.ChangeMetadata{
+				ID:          config.ChangeID("P135"),
+				Severity:    config.ChangeSeverityError,
+				Description: standardCoreEntityMessage("vault", "3.0"),
+				Resolution:  standardUpgradeMessage(versions300AndAbove),
```
GGabriele, line 850:
```suggestion
				Resolution:  standardUpgradeMessage("3.0"),
```

GGabriele: Otherwise this gets rendered like this:

`Please upgrade Kong Gateway to version '>= 3.0.0' or above.`

GGabriele:
I just added one minor last comment. I think we need to:
- make sure we can extend [this](https://github.com/Kong/koko/blob/41a67a99cc8a4c54c67eb212d57f55c72536e4ce/internal/grpc/proto/kong/admin/model/v1/vault.proto#L16) for other cases
- add an e2e test similar to [this](https://github.com/Kong/deck/commit/0df3f51cb93669eaf661051f292eb176923c0e88#diff-b09c72ddce91eda1770990ad2744a3867eb59e006e6842fc4d01b8d3ce5eec77R1901)

Other than that, this already looks great!

internal/server/kong/ws/config/compat/compatibility_table.go
EJKinger: fixed this

internal/server/kong/ws/config/compat/compatibility_table.go
```diff
@@ -834,6 +842,20 @@ var (
...
+		{
+			Metadata: config.ChangeMetadata{
+				ID:          config.ChangeID("P135"),
+				Severity:    config.ChangeSeverityError,
+				Description: standardCoreEntityMessage("vault", "3.0"),
+				Resolution:  standardUpgradeMessage("3.0"),
+			},
+			SemverRange: versionsPre300,
+			Update: config.ConfigTableUpdates{
+				Name:   "vault",
```
GGabriele, line 854:
```suggestion
				Name:   config.Vault.String(),
```
nit: for consistency with other changes

GGabriele (approved):
Minor nit, but other than that looks good. 

Great job!!!

internal/grpc/proto/kong/admin/model/v1/vault.proto
```diff
@@ -0,0 +1,19 @@
...
+option go_package = "github.com/kong/koko/internal/gen/grpc/kong/admin/model/v1;v1";
+
+message Vault {
+  string id = 1;
+  int32 created_at = 2;
+  int32 updated_at = 3;
+  string prefix = 4 [(google.api.field_behavior) = REQUIRED];
+  string name = 5 [(google.api.field_behavior) = REQUIRED];
+  string description = 6;
+  google.protobuf.Struct config = 7;
```
tjasko, line 17: Can we get this properly typed instead?
```protobuf
message Vault {
  message GCPSecretManager {
    string project_id = 1;
  }
  message AWSSecretsManager {
    string region = 1;
  }
  message HashiCorpVault {
    string host = 1;
    int32 port = 2;
    string protocol = 3;
    string mount = 4;
    string kv = 5;
    string token = 6;
  }
  message EnvironmentVariables {
    string prefix = 1;
  }

  oneof config {
    GCPSecretManager gcp_secret_manager = 7;
    AWSSecretsManager aws_secrets_manager = 8;
    HashiCorpVault hashicorp_vault = 9;
    EnvironmentVariables env = 10;
  }
}
```

internal/resource/vault_test.go
```diff
@@ -0,0 +1,232 @@
...
+	require.NotNil(t, r.Vault)
+}
+
+func TestVault_Type(t *testing.T) {
+	require.Equal(t, TypeVault, NewVault().Type())
+}
+
+func TestVault_ProcessDefaults(t *testing.T) {
+	vault := NewVault()
+	require.Nil(t, vault.ProcessDefaults(context.Background()))
```
tjasko, line 27: nit: `require.NoError` is better suited here.

internal/server/admin/vault_service.go
```diff
@@ -0,0 +1,140 @@
...
+	for _, object := range objects {
+		vault, ok := object.Resource().(*pb.Vault)
+		if !ok {
+			panic(fmt.Sprintf("expected type '%T' but got '%T'",
+				&pb.Vault{}, object.Resource()))
+		}
+		res = append(res, vault)
+	}
+	return res
+}
```
tjasko, line 140: nit: We know the length:
```suggestion
func vaultsFromObjects(objects []model.Object) []*pb. Vault {
	res := make([]*pb. Vault, len(objects))
	for i, object := range objects {
		var ok bool
		if res[i], ok = object.Resource().(*pb. Vault); !ok {
			panic(fmt.Sprintf("expected type '%T' but got '%T'",
				&pb. Vault{}, object.Resource()))
		}
	}
	return res
}
```

internal/server/kong/ws/config/compat/compatibility_table.go
```diff
@@ -834,6 +842,20 @@ var (
 				},
 			},
 		},
+		{
+			Metadata: config.ChangeMetadata{
+				ID:          config.ChangeID("P135"),
```
tjasko, line 847: We'll want to get in a test for this.

internal/test/e2e/basic_test.go
```diff
@@ -564,6 +564,37 @@ func TestSNISync(t *testing.T) {
...
+		Name:   "env",
+		Prefix: "test-vault-1",
+	}
+	c := httpexpect.New(t, "http://localhost:3000")
+	c.POST("/v1/vaults").WithJSON(vault).Expect().Status(http.StatusCreated)
+
+	dpCleanup := run.KongDP(kong.GetKongConfForShared())
+	defer dpCleanup()
+
+	require.Nil(t, util.WaitForKongPort(t, 8001))
```
tjasko, line 584: nit: `require.NoError` is better suited here.

internal/grpc/proto/kong/admin/model/v1/vault.proto
EJKinger: In order to match the existing api, I think we'll have to go with something like:

```protobuf
message Vault {
  string id = 1;
  int32 created_at = 2;
  int32 updated_at = 3;
  string prefix = 4 [(google.api.field_behavior) = REQUIRED];
  string name = 5 [(google.api.field_behavior) = REQUIRED];
  string description = 6;

  message Config {
    string prefix = 1;
    string region = 2;
    string project_id = 3;
    string host = 4;
    int32 port = 5;
    string protocol = 6;
    string mount = 7;
    string kv = 8;
    string token = 9;
  }

  Config config =  7;
  repeated string tags = 8;
}
```

internal/resource/vault_test.go
EJKinger: fixed

internal/server/admin/vault_service.go
EJKinger: fixed

internal/test/e2e/basic_test.go
EJKinger: fixed

internal/server/kong/ws/config/compat/compatibility_table.go
EJKinger: Good catch on this test, I'm pretty sure that core entity deletion hasn't actually been implemented.

mikefero: It is, you will just need to add a new vaults type; see [here](https://github.com/Kong/koko/blob/738960912041441f287456f1fdbeefe58da43cfc/internal/server/kong/ws/config/version_compatibility.go#L32-L51). From here you will want to add an [entry to the compatibility table along with change metadata information](https://github.com/Kong/koko/blob/738960912041441f287456f1fdbeefe58da43cfc/internal/server/kong/ws/config/compat/compatibility_table.go#L86) and then [this entity removal loop](https://github.com/Kong/koko/blob/738960912041441f287456f1fdbeefe58da43cfc/internal/server/kong/ws/config/version_compatibility.go#L521-L542) will take over the action.

EJKinger: @mikefero this isn't a field removal, it's removing the entire vaults entity.

EJKinger: https://github.com/Kong/koko/pull/464/files#diff-ca9bb92e11fab1c215299a0598bbd953852f976044f0bc8cd18cd8a0b9912d50R264-R302

Some extra eyes on this part would be great, I am not super familiar with this part of the code, but I think I have it patched together.

internal/server/kong/ws/config/version_compatibility.go
```diff
@@ -260,6 +261,45 @@ func (vc *WSVersionCompatibility) processConfigTableUpdates(uncompressedPayload
...
+	processedPayload, err := sjson.Delete(processedPayload, fmt.Sprintf("config_table.%ss", entityType))
+	if err != nil {
+		vc.logger.With(zap.String("entity", entityType)).
+			With(zap.String("data-plane", dataPlaneVersionStr)).
+			With(zap.Error(err)).
+			Error("entity was not removed from configuration")
+	} else {
+		vc.logger.With(zap.String("entity", entityType)).
+			With(zap.String("data-plane", dataPlaneVersionStr)).
+			Warn("removing %ss entity which is incompatible with data plane")
```
GGabriele, line 297:
```suggestion
			Warn("removing entity which is incompatible with data plane")
```

No need to add the entity type here, as that's already part of the logger `zap.String("entity", entityType)`

internal/server/kong/ws/config/compat/compatibility_table.go
GGabriele: Nice, I just added some comments.

On top of those:
- can you add some test for this in `version_compatibility_test.go`? (as pointed out by Taylor)
- can you also add some e2e test [here](https://github.com/Kong/koko/blob/main/internal/test/e2e/version_compatibility_test.go) to make sure this entity is remove for DP < 3.0.0?

mikefero: > @mikefero this isn't a field removal, it's removing the entire vaults entity.

:man_facepalming: You are 100% correct.

In that case we replicate [removePlugin](https://github.com/Kong/koko/blob/738960912041441f287456f1fdbeefe58da43cfc/internal/server/kong/ws/config/version_compatibility.go#L266-L306) functionality except do it for entities. I would say that is kind of feature creep for this PR and would be better suited in a separate one.

internal/resource/vault_test.go
EJKinger: @GGabriele these are included.

internal/server/kong/ws/config/compat/compatibility_table.go
EJKinger: #483

internal/server/kong/ws/config/version_compatibility.go
```diff
@@ -260,6 +261,45 @@ func (vc *WSVersionCompatibility) processConfigTableUpdates(uncompressedPayload
...
 
+func (vc *WSVersionCompatibility) removeCoreEntity(
+	processedPayload string,
+	entityType string,
+	dataPlaneVersionStr string,
+	changeID ChangeID,
+	tracker *ChangeTracker,
+) string {
+	entity := gjson.Get(processedPayload, fmt.Sprintf("config_table.%ss", entityType))
+	if entity.IsArray() {
```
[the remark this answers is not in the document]
EJKinger: I made these changes in #483

internal/server/kong/ws/config/version_compatibility.go
```diff
@@ -260,6 +261,45 @@ func (vc *WSVersionCompatibility) processConfigTableUpdates(uncompressedPayload
...
 }
 
+func (vc *WSVersionCompatibility) removeCoreEntity(
+	processedPayload string,
+	entityType string,
+	dataPlaneVersionStr string,
+	changeID ChangeID,
+	tracker *ChangeTracker,
+) string {
+	entity := gjson.Get(processedPayload, fmt.Sprintf("config_table.%ss", entityType))
```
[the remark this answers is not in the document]
EJKinger: I made these changes in #483

internal/server/kong/ws/config/version_compatibility.go
EJKinger: I made these changes in #483

GGabriele (approved):
Great work!!!

internal/test/e2e/basic_test.go
```diff
@@ -564,6 +564,89 @@ func TestSNISync(t *testing.T) {
...
+func TestVaultSync(t *testing.T) {
+	// ensure that Vaults can be synced to Kong gateway
+	cleanup := run.Koko(t)
+
+	defer cleanup()
+
+	vault := &v1.Vault{
+		Id:     uuid.NewString(),
+		Name:   "env",
+		Prefix: "test-vault-1",
```
GGabriele, line 576: Please add a `Config` here.
Right now the sync to DPs won't work and this test doesn't catch it.

GGabriele (changes requested):
The current way this is implemented cannot sync configuration down to DPs:

```
Error: rr: 2022/10/20 08:31:22 [error] 1109#0: *6 [lua] wrpc_data_plane.lua:180: [wrpc-clustering] unable to update running config: bad config received from control plane in 'vaults':
dp-stderr:   - in entry 1 of 'vaults':
dp-stderr:     in 'config':
dp-stderr:       in 'env': unknown field, context: ngx.timer
```

internal/test/e2e/basic_test.go
GGabriele: The error looks like this:

```
Error: rr: 2022/10/20 08:31:22 [error] 1109#0: *6 [lua] wrpc_data_plane.lua:180: [wrpc-clustering] unable to update running config: bad config received from control plane in 'vaults':
dp-stderr:   - in entry 1 of 'vaults':
dp-stderr:     in 'config':
dp-stderr:       in 'env': unknown field, context: ngx.timer
```

This is happening because we are modelling the Vault object differently than what the DP expects.

internal/test/e2e/basic_test.go
```diff
@@ -564,6 +564,89 @@ func TestSNISync(t *testing.T) {
...
+	require.NoError(t, util.WaitForKongPort(t, 8001))
+
+	// kongClient
+	kongClient.RunWhenKong(t, ">= 3.0.0")
+
+	expectedConfig := &v1.TestingConfig{
+		Vaults: []*v1.Vault{vault},
+	}
+	util.WaitFunc(t, func() error {
+		err := util.EnsureConfig(expectedConfig)
```
GGabriele, line 593: Once you do [this](https://github.com/Kong/koko/pull/464/files#r1000443742), `EnsureConfig` is no longer usable.
You'd need to create something that accepts Kong structures:

```
func EnsureKongConfig(expectedConfig KongConfig) error {
	gotConfig, err := fetchKongConfig()
	if err != nil {
		return fmt.Errorf("fetching kong config: %v", err)
	}
	err = JSONSubset(expectedConfig, gotConfig)
	return err
}
```

internal/grpc/proto/kong/admin/model/v1/vault.proto
EJKinger: I'l update this test to catch this.

internal/server/kong/ws/config/vault.go
```diff
@@ -0,0 +1,77 @@
...
+		}
+		delete(m, "updated_at")
+		res = append(res, flattenVaultConfig(m))
+	}
+	config["vaults"] = res
+	return nil
+}
+
+func flattenVaultConfig(m Map) Map {
+	if _, ok := m["config"]; ok {
```
GGabriele, line 59:
```suggestion
	if _, ok := m["config"]; !ok {
	    continue
	}    
```
Doing something like this would require less nesting, making core more readable

GGabriele (dismissed):

GGabriele:
+1!

Please wait for another +1 from @tjasko or @mikefero

GGabriele (approved):

internal/test/e2e/vault_test.go
```diff
@@ -0,0 +1,265 @@
...
+		Id:     uuid.NewString(),
+		Name:   "env",
+		Prefix: "my-env-vault",
+		Config: &v1.Vault_Config{
+			Config: &v1.Vault_Config_Env{
+				Env: &v1.Vault_EnvConfig{
+					Prefix: "KONG_MY_SECRET_",
+				},
+			},
+		},
```
tjasko, line 112: We should test that creating a vault without the config set fails, right?

Asking about this because:
https://github.com/Kong/koko/pull/464/files#diff-c394f5355ba1aca720f6e623a017d131930f0ee830a9215f6e6961c29c0e3f5eR110

internal/grpc/proto/kong/admin/model/v1/vault.proto
```diff
@@ -0,0 +1,44 @@
...
+  message HcvConfig {
+    string host = 1;
+    int32 port = 2;
+    string protocol = 3;
+    string mount = 4;
+    string kv = 5;
+    string token = 6;
+  }
+  message Config {
+    oneof config {
```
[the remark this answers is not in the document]
EJKinger: Without nesting this, the result was that each of these configuration blocks had their own top level key.  I wanted the keys to be nested under the top level "config" key to more closely match the kong api.

internal/resource/vault.go
```diff
@@ -0,0 +1,147 @@
...
+						},
+					},
+				},
+			},
+			"description": {Type: "string"},
+			"tags":        typedefs.Tags,
+			"created_at":  typedefs.UnixEpoch,
+			"updated_at":  typedefs.UnixEpoch,
+		},
+		Required: []string{
```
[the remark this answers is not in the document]
EJKinger: No, config isn't required.

internal/test/util/kong_config.go
```diff
@@ -69,6 +79,12 @@ func fetchKongConfig() (KongConfig, error) {
 	if err != nil {
 		return KongConfig{}, fmt.Errorf("fetch SNIs: %v", err)
 	}
+	// TODO (ejkinger): handle this based on version compatibility
+	vaults, _ := client.Vaults.ListAll(ctx)
```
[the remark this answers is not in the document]
EJKinger: fixed this.

GGabriele (dismissed):

internal/grpc/proto/kong/admin/model/v1/vault.proto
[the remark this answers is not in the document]
tjasko: Ah, got it, that's completely fair.

internal/test/e2e/vault_test.go
tjasko: Ah interesting, alrighty then.

internal/test/util/kong_config.go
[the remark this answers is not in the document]
tjasko: We'll still need to re-assign `err` here, as it's missing:
```suggestion
	vaults, err := client.Vaults.ListAll(ctx)
```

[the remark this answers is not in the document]
tjasko: We'll still need to re-assign `err` here, as it's missing:
```suggestion
	vaults, err := client.Vaults.ListAll(ctx)
```

internal/test/util/kong_config.go
```diff
@@ -69,6 +78,10 @@ func fetchKongConfig() (KongConfig, error) {
 	if err != nil {
 		return KongConfig{}, fmt.Errorf("fetch SNIs: %v", err)
 	}
+	vaults, err := client.Vaults.ListAll(ctx)
+	if err != nil {
+		return KongConfig{}, fmt.Errorf("fetch Vaults: %v", err)
```
GGabriele, line 83: Only newer DPs have the `/vaults` endpoint, so returning an error here would make all tests to fail ([example](https://github.com/Kong/koko/actions/runs/3340568631/jobs/5530817665))

EJKinger: I was thinking that since you had added `kongClient.RunWhenKong(t, ">= 3.0.0")` to the vault tests that this wouldn't be a problem, but I forgot about all of the other tests that use this.

internal/test/util/kong_config.go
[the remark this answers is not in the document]
tjasko: Confused, why don't we care about the error here? An API error can have this correctly return in the event of their being one.

GGabriele (approved):

internal/test/util/kong_config.go
```diff
@@ -69,6 +78,9 @@ func fetchKongConfig() (KongConfig, error) {
 	if err != nil {
 		return KongConfig{}, fmt.Errorf("fetch SNIs: %v", err)
 	}
+	vaults, _ := client.Vaults.ListAll(ctx)
```
GGabriele, line 81: What you can do is checking if the error code and ignore 404s.

tjasko: Yeah, this [comment](https://github.com/Kong/koko/pull/464/files#r1003665778) talks about it too. I ran all the E2E tests and this call isn't expected to ever error, as it's just listing vaults.

What would you said about just doing the below? If so, don't mind making that change.
```suggestion
	vaults, err := client.Vaults.ListAll(ctx)
	if err != nil {
		return KongConfig{}, fmt.Errorf("unable to fetch vaults: %w", err)
	}
```

GGabriele: This is actually expected to error out with 404s for DPs not supporting Vaults, so pre-3.0

tjasko: Derp, right... will get in the proper fix for this. Completely sidestepped my mind this was calling the Kong API directly.

tjasko: Derp, right... will get in the proper fix for this.

tjasko (approved):

PR #468, opened 2022-10-10
fix: correct compose type

rainest:
Use a "true" string instead of a boolean in compose environment.

Makes YAML less unhappy about:

```
 $ docker-compose up     
ERROR: The Compose file './docker-compose.yml' is invalid because:
services.koko.environment.KOKO_DATABASE_SQLITE_IN_MEMORY contains true, which is an invalid type, it should be a string, number, or a null
```

### DIFF
--- a/docs/assets/docker-compose.yml
+++ b/docs/assets/docker-compose.yml
@@ -9,7 +9,7 @@ services:
     image: "kong/koko:latest"
     command: "serve"
     environment:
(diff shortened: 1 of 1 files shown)


GGabriele (approved):

tjasko (approved):

rainest:
@GGabriele @tjasko I have satisfied the CLA bot, not sure if it would notify that CI's now happy on its own.

tjasko:
Thanks @rainest, merging this in.

PR #469, opened 2022-10-10
fix: conditionally ignore empty value updates

omegabytes:
Fixes a bug where if either a pre-function or post-function plugin config contained config.access
and was < v3.0.0 with a downstream dp > v3.0.0, the access field is overwritten with the empty array
and is rejected by the dp.

Generally: updates the processing performed on plugin configs during version compatibility updates.
Previously, an empty array value would overwrite an existing target array value when performing the update.
This is an edge case as the target value being overwritten is on a field that does not exist
for the control-plane version but does exist for the data-plane config.

More concretely:
* Version < v3.0.0
  * creates a default value functions: [] when functions is undefined
  * access field does not exist (but can be defined by user)
* Version > v3.0.0
  * Creates an access field and copies the value contained in functions
  * Removes functions field from config
* Edge-case: config version < 3.0.0 contains access: ["some-value"] and functions is undefined
  * Under-the-hood, functions is set to default value functions: []
  * Since functions field exists, version compat overwrites existing access with empty value
  * Downstream > 3.0.0 dp rejects config validation as access must not be empty

### DIFF
--- a/internal/server/kong/ws/config/compat/compatibility_table.go
+++ b/internal/server/kong/ws/config/compat/compatibility_table.go
@@ -603,6 +603,7 @@ var (
 							{
 								Field:          "access",
 								ValueFromField: "functions",
+								IgnoreEmpty:    true,
 							},
 							{
 								Field: "functions",
@@ -636,6 +637,7 @@ var (
 							{
 								Field:          "access",
 								ValueFromField: "functions",
+								IgnoreEmpty:    true,
 							},
 							{
 								Field: "functions",
--- a/internal/server/kong/ws/config/compat/compatibility_table.go
+++ b/internal/server/kong/ws/config/compat/compatibility_table.go
@@ -601,9 +601,9 @@ var (
 						Condition: "functions",
 						Updates: []config.ConfigTableFieldUpdate{
 							{
-								Field:          "access",
-								ValueFromField: "functions",
-								IgnoreEmpty:    true,
+								Field:            "access",
+								ValueFromField:   "functions",
+								FieldMustBeEmpty: true,
 							},
 							{
 								Field: "functions",
@@ -635,9 +635,9 @@ var (
 						Condition: "functions",
 						Updates: []config.ConfigTableFieldUpdate{
 							{
-								Field:          "access",
-								ValueFromField: "functions",
-								IgnoreEmpty:    true,
+								Field:            "access",
+								ValueFromField:   "functions",
+								FieldMustBeEmpty: true,
 							},
 							{
 								Field: "functions",
--- a/internal/server/kong/ws/config/compat/compatibility_table.go
+++ b/internal/server/kong/ws/config/compat/compatibility_table.go
@@ -836,6 +836,42 @@ var (
 				},
 			},
 		},
+		{
+			Metadata: config.ChangeMetadata{
+				ID:       config.ChangeID("P135"),
+				Severity: config.ChangeSeverityWarning,
+				Description: "For the 'pre-function' plugin, " +
+					"'config.functions' field has been used. " +
+					"This field is deprecated and it is no longer supported " +
+					"in Kong Gateway versions >= 3.0.",
+				Resolution: "Please update the plugin configuration to use " +
+					"'config.access' field in place of 'config.functions' field",
+			},
+			SemverRange: versions300AndAbove,
+			Update: config.ConfigTableUpdates{
+				Name:         "pre-function",
+				Type:         config.Plugin,
+				RemoveFields: []string{"functions"},
+			},
+		},
+		{
+			Metadata: config.ChangeMetadata{
+				ID:       config.ChangeID("P136"),
+				Severity: config.ChangeSeverityWarning,
+				Description: "For the 'post-function' plugin, " +
+					"'config.functions' field has been used. " +
+					"This field is deprecated and it is no longer supported " +
+					"in Kong Gateway versions >= 3.0.",
+				Resolution: "Please update the plugin configuration to use " +
+					"'config.access' field in place of 'config.functions' field",
+			},
+			SemverRange: versions300AndAbove,
+			Update: config.ConfigTableUpdates{
+				Name:         "post-function",
+				Type:         config.Plugin,
+				RemoveFields: []string{"functions"},
+			},
+		},
 	}
 )
 
--- a/internal/server/kong/ws/config/compat/compatibility_table.go
+++ b/internal/server/kong/ws/config/compat/compatibility_table.go
@@ -595,6 +595,11 @@ var (
 			Update: config.ConfigTableUpdates{
 				Name: "pre-function",
 				Type: config.Plugin,
+				DisableChangeTracking: func(rawJSON string) bool {
+					// do not emit change if functions is set to default value (empty array)
+					plugin := gjson.Parse(rawJSON)
+					return config.ValueIsEmpty(plugin.Get("config.functions"))
+				},
 				FieldUpdates: []config.ConfigTableFieldCondition{
 					{
 						Field:     "functions",
@@ -629,6 +634,11 @@ var (
 			Update: config.ConfigTableUpdates{
 				Name: "post-function",
 				Type: config.Plugin,
+				DisableChangeTracking: func(rawJSON string) bool {
+					// do not emit change if functions is set to default value (empty array)
+					plugin := gjson.Parse(rawJSON)
+					return config.ValueIsEmpty(plugin.Get("config.functions"))
+				},
 				FieldUpdates: []config.ConfigTableFieldCondition{
 					{
 						Field:     "functions",
--- a/internal/server/kong/ws/config/compat/compatibility_table.go
+++ b/internal/server/kong/ws/config/compat/compatibility_table.go
@@ -836,48 +836,6 @@ var (
 				},
 			},
 		},
-		{
-			Metadata: config.ChangeMetadata{
-				ID:       config.ChangeID("P135"),
-				Severity: config.ChangeSeverityError,
-				Description: standardPluginFieldsMessage("pre-function",
-					[]string{"functions"}, "3.0", true),
-				Resolution: "Please update the plugin configuration to use " +
-					"'config.access' field in place of 'config.functions' field",
-			},
-			SemverRange: versions300AndAbove,
-			Update: config.ConfigTableUpdates{
-				Name:         "pre-function",
-				Type:         config.Plugin,
-				RemoveFields: []string{"functions"},
-				DisableChangeTracking: func(rawJSON string) bool {
-					// do not emit change if functions is set to default value (empty array)
-					plugin := gjson.Parse(rawJSON)
-					return len(plugin.Get("config.functions").Array()) == 0
-				},
-			},
-		},
-		{
-			Metadata: config.ChangeMetadata{
-				ID:       config.ChangeID("P136"),
-				Severity: config.ChangeSeverityError,
-				Description: standardPluginFieldsMessage("post-function",
-					[]string{"functions"}, "3.0", true),
-				Resolution: "Please update the plugin configuration to use " +
-					"'config.access' field in place of 'config.functions' field",
-			},
-			SemverRange: versions300AndAbove,
-			Update: config.ConfigTableUpdates{
-				Name:         "post-function",
-				Type:         config.Plugin,
-				RemoveFields: []string{"functions"},
-				DisableChangeTracking: func(rawJSON string) bool {
-					// do not emit change if functions is set to default value (empty array)
-					plugin := gjson.Parse(rawJSON)
-					return len(plugin.Get("config.functions").Array()) == 0
-				},
-			},
-		},
 	}
 )
 
--- a/internal/server/kong/ws/config/compat/compatibility_table.go
+++ b/internal/server/kong/ws/config/compat/compatibility_table.go
@@ -839,7 +839,7 @@ var (
 		{
 			Metadata: config.ChangeMetadata{
 				ID:       config.ChangeID("P135"),
-				Severity: config.ChangeSeverityWarning,
+				Severity: config.ChangeSeverityError,
 				Description: standardPluginFieldsMessage("pre-function",
 					[]string{"functions"}, "3.0", true),
 				Resolution: "Please update the plugin configuration to use " +
@@ -850,12 +850,17 @@ var (
 				Name:         "pre-function",
 				Type:         config.Plugin,
 				RemoveFields: []string{"functions"},
+				DisableChangeTracking: func(rawJSON string) bool {
+					// do not emit change if functions is set to default value (empty array)
+					plugin := gjson.Parse(rawJSON)
+					return len(plugin.Get("config.functions").Array()) == 0
+				},
 			},
 		},
 		{
 			Metadata: config.ChangeMetadata{
 				ID:       config.ChangeID("P136"),
-				Severity: config.ChangeSeverityWarning,
+				Severity: config.ChangeSeverityError,
 				Description: standardPluginFieldsMessage("post-function",
 					[]string{"functions"}, "3.0", true),
 				Resolution: "Please update the plugin configuration to use " +
@@ -866,6 +871,11 @@ var (
 				Name:         "post-function",
 				Type:         config.Plugin,
 				RemoveFields: []string{"functions"},
+				DisableChangeTracking: func(rawJSON string) bool {
+					// do not emit change if functions is set to default value (empty array)
+					plugin := gjson.Parse(rawJSON)
+					return len(plugin.Get("config.functions").Array()) == 0
+				},
 			},
 		},
 	}
--- a/internal/server/kong/ws/config/compat/compatibility_table.go
+++ b/internal/server/kong/ws/config/compat/compatibility_table.go
@@ -840,10 +840,8 @@ var (
 			Metadata: config.ChangeMetadata{
 				ID:       config.ChangeID("P135"),
 				Severity: config.ChangeSeverityWarning,
-				Description: "For the 'pre-function' plugin, " +
-					"'config.functions' field has been used. " +
-					"This field is deprecated and it is no longer supported " +
-					"in Kong Gateway versions >= 3.0.",
+				Description: standardPluginFieldsMessage("pre-function",
+					[]string{"functions"}, "3.0", true),
 				Resolution: "Please update the plugin configuration to use " +
 					"'config.access' field in place of 'config.functions' field",
 			},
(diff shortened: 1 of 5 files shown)


internal/server/kong/ws/config/version_compatibility_test.go
```diff
@@ -5451,6 +5451,79 @@ func TestVersionCompatibility_ProcessConfigTableUpdates(t *testing.T) {
...
+						Resources: []ResourceInfo{
+							{
+								Type: "plugin",
+								ID:   "759c0d3a-bc3d-4ccc-8d4d-f92de95c1f1a",
+							},
+						},
+					},
+				},
+			},
+		},
```
omegabytes, line 5526: We detect the presence of the default-value `functions: []` wether the user added it or it was automatically created as the default value, then notify the user of the change. Changes to this behavior require more advanced handling.

internal/server/kong/ws/config/version_compatibility.go
```diff
@@ -678,3 +685,27 @@ func shouldTrackChange(updates ConfigTableUpdates, entityJSON string) bool {
 	}
 	return !updates.DisableChangeTracking(entityJSON)
 }
+
+// fieldIsEmpty checks if a given field is one of:
+//   - Object with no nested items
```
GGabriele, line 690:
```suggestion
//   - Object with no items
```

internal/server/kong/ws/config/version_compatibility.go
```diff
@@ -678,3 +685,27 @@ func shouldTrackChange(updates ConfigTableUpdates, entityJSON string) bool {
...
+//   - Type is null
```
GGabriele, line 691:
```suggestion
//   - Value is null
```

internal/server/kong/ws/config/version_compatibility.go
```diff
@@ -678,3 +685,27 @@ func shouldTrackChange(updates ConfigTableUpdates, entityJSON string) bool {
...
+//   - Zero length array
+//   - Zero length string
```
GGabriele, line 693:
```suggestion
//   - String is empty
```
Technically, you're not checking the length

internal/server/kong/ws/config/version_compatibility_test.go
```diff
@@ -5451,6 +5451,408 @@ func TestVersionCompatibility_ProcessConfigTableUpdates(t *testing.T) {
...
+						Type: Plugin,
+						RemoveFields: []string{
+							"plugin_field_1",
+						},
+						ChangeID: "T102",
+					},
+				},
+			},
+			uncompressedPayload: `
+		{
```
GGabriele, line 5490: Can you please make sure this is aligned with the other code?

internal/server/kong/ws/config/compat/compatibility_table.go
```diff
@@ -834,6 +836,42 @@ var (
 				},
 			},
 		},
+		{
+			Metadata: config.ChangeMetadata{
```
GGabriele, line 840: Why are these needed since we already have `P122` and `P123`? 

I think this can lead to some weird edge cases like this:
- `config.functions` is set to a non-null value
- `P135` is processed, removing the `functions` field
- `P122` doesn't get processed because `functions` is not set anymore, therefore `access` is not correctly populated

Right now there is no guarantee that updates are [processed in order](https://github.com/Kong/koko/pull/414)

internal/server/kong/ws/config/compat/compatibility_table.go
```diff
@@ -834,6 +836,42 @@ var (
...
+				ID:       config.ChangeID("P135"),
+				Severity: config.ChangeSeverityWarning,
+				Description: "For the 'pre-function' plugin, " +
```
GGabriele, line 843: Note: you could use `standardPluginFieldsMessage` for this.

internal/server/kong/ws/config/compat/compatibility_table.go
GGabriele: Oh I see why you're doing this.

Still, I think this can lead to some weird issue. What about short-circuiting the `ValueFromField` but not the field removal? This way we would keep the changes limited to how the `ValueFromField` is used, which is where the bug resides.

WDYT? (cc: @mikefero )

GGabriele:
Added some comments and opinions. Just a couple of more points, other than those:
- can you please make sure we have an e2e test case covering this? 
- can you please include any useful description of this patch in the commit message (like you did in the PR description)?

internal/server/kong/ws/config/compat/compatibility_table.go
GGabriele: After talking with @mikefero I think we are good on this front: the ordering issue **is** indeed an issue, but not for this specific case. 

Can you please make sure we have a ticket to track this though?

internal/server/kong/ws/config/compat/compatibility_table.go
```diff
@@ -636,6 +637,7 @@ var (
 							{
 								Field:          "access",
 								ValueFromField: "functions",
+								IgnoreEmpty:    true,
```
mikefero, line 640: `IgnoreEmpty` in this case doesn't make sense; this should be when the `Field` is **not** empty. Another way of putting this is when `config.access` has a value (e.g. not empty) then the value should not be copied and the field update process for this change should be short circuited.

internal/test/e2e/plugins.go
```diff
@@ -254,6 +275,27 @@ var VersionCompatibilityOSSPluginConfigurationTests = []VersionCompatibilityPlug
...
+					Field: "access",
+					Value: []string{
+						"kong.log.err('Hello Koko!')",
+					},
+				},
+			},
+		},
+		ConfigureForService: true,
+		ConfigureForRoute:   true,
+	},
```
mikefero, line 298: These two new additions to plugins.go for `post-function` and `pre-function` won't work as the plugins will already be configured. This end-to-end test is required, but it will need to be in it's own test; e.g. `TestVersionCompatibility_EnsureAccessFieldAreNotOverriddenWithEmptyFields` and only if the data plane >= 3.0.

internal/server/kong/ws/config/compat/compatibility_table.go
omegabytes: I can update it, but we talked about ignoring fields with empty values, not wholesale ignoring existing fields when the target field exists already.

mikefero: Fair point and the described task should have better reflected this. Looking at the issue more closely we wouldn't want to have configured fields (e.g. not empty) be overwritten regardless of what another field has in it configuration.

For example using the `pre-function` serverless plugin, if the plugin is configured the following way:

```json
{
  "access": [
    "kong.log.err('access')"
  ]
  "functions": [
    "kong.log.err('functions')"
  ]
}
```

In this example we would not want to override `config.access` just because `config.functions` is not empty; whatever is assigned is `config.access` in this instance should remain. Does that clear things up?

omegabytes: Yes, thank you. I can think of a few of edge cases, like what should the expected behavior be in this case:
```
{
  "access": []
  "functions": [
    "kong.log.err('functions')"
  ]
}
```
I don't think we can answer that generally for all present and future plugin configs.

mikefero: Nope and that is why the flag will be beneficial to have in this case. For this case though if `config.access` is "empty" we want `config.functions` to be copied into `config.access` for >= 3.0.

internal/server/kong/ws/config/compat/compatibility_table.go
omegabytes: I'm not really understanding, unfortunately. I'm pushing some updates to how this works based on review feedback which will impact expected behavior so do you mind revisiting on the next review pass?

internal/server/kong/ws/config/version_compatibility.go
```diff
@@ -386,6 +389,10 @@ func (vc *WSVersionCompatibility) processPluginUpdates(payload string,
 					conditionField := fmt.Sprintf("[@this].#(config.%s)", update.Condition)
 					if gjson.Get(updatedRaw, conditionField).Exists() {
 						for _, fieldUpdate := range update.Updates {
+							if fieldUpdate.IgnoreEmpty && fieldIsEmpty(gjson.Get(updatedRaw, configField)) {
+								break
+							}
```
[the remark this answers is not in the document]
mikefero: :100: We are not using it currently, but the condition will be the same and the unit tests for `Service`, `Route`, and `Upstream` `UpdateType`s can be handled in a loop for the unit tests.

Here is some pseudo code

```go
tests := []struct{
.
.
.
}

entityTypes := []UpdateType{
  Service,
  Route,
  Upstream,
}
for _, entityType := range entityTypes {
  // Handle not overriding a configured field
  tests = append(tests, struct{<tests-struct-format})
  // Handle overriding an empty field
  tests = append(tests, struct{<tests-struct-format})
}
```

internal/server/kong/ws/config/version_compatibility.go
```diff
@@ -678,3 +685,27 @@ func shouldTrackChange(updates ConfigTableUpdates, entityJSON string) bool {
...
+	}
+
+	if field.Value() == nil {
+		return true
+	}
+
+	if field.String() == "" {
+		return true
+	}
+	return false
```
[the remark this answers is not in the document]
omegabytes: [8d2312c](https://github.com/Kong/koko/pull/469/commits/8d2312c06d8ab1a1e8170a4a8b0c537560d49c21)

`fieldValue: {[]}` isn't valid JSON, so the test case would look like: `fieldValue: { "foo": []}`. This is not considered empty. I will need to add handling that performs the empty check on every nested element. Is that what we want?

Should we consider this empty?
```
{
    "items": [
        {
            "sub1": [
                {
                    "sub2": {}
                }
            ]
        }
    ]
}
```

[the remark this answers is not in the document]
mikefero: > fieldValue: {[]} isn't valid JSON

good call; told you it wasn't tested :smile:

> Should we consider this empty?

No, if there is any field in the object block like `{ "foo": [] }` that should not be considered empty as it has a `key` with an "empty" `value`.

internal/test/e2e/version_compatibility_test.go
```diff
@@ -207,6 +207,155 @@ func TestVersionCompatibility_PluginFieldUpdates(t *testing.T) {
...
+			Config:    &config,
+			Enabled:   wrapperspb.Bool(true),
+			Protocols: []string{"http", "https"},
+		}
+
+		pluginBytes, err := json.ProtoJSONMarshal(p)
+		require.NoError(t, err)
+		res := admin.POST("/v1/plugins").WithBytes(pluginBytes).Expect()
+		res.Status(http.StatusCreated)
+		require.Equal(t, http.StatusCreated, res.Raw().StatusCode)
```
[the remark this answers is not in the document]
omegabytes: It would be faster if we caught and killed tests with invalid configs. This res.Status call doesn't cause the test to fail and halt.

internal/server/kong/ws/config/version_compatibility.go
```diff
@@ -570,6 +580,12 @@ func (vc *WSVersionCompatibility) processCoreEntityUpdates(payload string,
 				if gjson.Get(updatedRaw, conditionField).Exists() {
 					for _, fieldUpdate := range update.Updates {
 						conditionUpdate := fieldUpdate.Field
+						// Ensure the original field is not empty if specified; do not overwrite
+						if fieldUpdate.FieldMustBeEmpty && !valueIsEmpty(gjson.Get(updatedRaw, conditionUpdate)) {
+							// Since this is a copy function from another field and the current field is already
+							// configured the entire field update process should short circuit
+							break
```
GGabriele, line 587: I guess this should be `continue` like in line 398?

internal/test/e2e/version_compatibility_test.go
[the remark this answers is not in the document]
GGabriele: Then please change it in a way that the test fails fast in such case.

internal/server/kong/ws/config/version_compatibility.go
mikefero: Switching from `break` to `continue` in both of these should update the comments (and update the doc string for `FieldMustBeEmpty`.

internal/server/kong/ws/config/compat/compatibility_table.go
```diff
@@ -634,8 +635,9 @@ var (
 						Condition: "functions",
 						Updates: []config.ConfigTableFieldUpdate{
 							{
-								Field:          "access",
-								ValueFromField: "functions",
+								Field:            "access",
+								ValueFromField:   "functions",
+								FieldMustBeEmpty: true,
```
mikefero, line 640: Since we are switching from `break` to `continue` to ensure the ordering doesn't mess things up between removal and updates we should re-add:

```
DisableChangeTracking: func(rawJSON string) bool {
	// do not emit change if functions is set to default value (empty array)
	plugin := gjson.Parse(rawJSON)
	return len(plugin.Get("config.plugin_field_1").Array()) == 0
},
```

internal/test/e2e/version_compatibility_test.go
```diff
@@ -273,6 +273,152 @@ func TestVersionCompatibility_PluginFieldUpdates(t *testing.T) {
...
 }
 
+func TestVersionCompatibility_EnsureTargetFieldsAreNotOverridden(t *testing.T) {
+	cleanup := run.Koko(t)
+	defer cleanup()
+
+	dpCleanup := run.KongDP(kong.GetKongConfForShared())
+	defer dpCleanup()
+	require.NoError(t, util.WaitForKong(t))
+	require.NoError(t, util.WaitForKongAdminAPI(t))
```
mikefero, line 283: We should ensure this tests only runs for 3.0.0+.

```suggestion
	require.NoError(t, util.WaitForKongAdminAPI(t))
	
	kongClient.RunWhenKong(t, ">=3.0.0")
```

internal/server/kong/ws/config/version_compatibility.go
omegabytes: Thanks for catching!

internal/server/kong/ws/config/compat/compatibility_table.go
omegabytes: Switching `break` to `continue` here also requires switching from `break` to `continue` [here](https://github.com/Kong/koko/blob/main/internal/server/kong/ws/config/version_compatibility.go#L610), which causes [this test](https://github.com/Kong/koko/blob/main/internal/server/kong/ws/config/version_compatibility_test.go#L5409) to fail. Not switching means [ his test](https://github.com/Kong/koko/blob/18e185dd9efd41e76298224671f9716a486c925b/internal/server/kong/ws/config/version_compatibility_test.go#L5845-L5902) case will fail:
```
{
			name: "field updates do not create new fields when ignoring empty array values",
			configTableUpdates: map[string][]ConfigTableUpdates{
				">= 3.0.0": {
					{
						Name: "plugin_1",
						Type: Plugin,
						DisableChangeTracking: func(rawJSON string) bool {
							plugin := gjson.Parse(rawJSON)
							return ValueIsEmpty(plugin.Get("config.plugin_field_1"))
						},
						FieldUpdates: []ConfigTableFieldCondition{
							{
								Field:     "plugin_field_1",
								Condition: "plugin_field_1",
								Updates: []ConfigTableFieldUpdate{
									{
										Field:            "plugin_field_2",
										ValueFromField:   "plugin_field_1",
										FieldMustBeEmpty: true,
									},
									{
										Field: "plugin_field_1",
									},
								},
							},
						},
						ChangeID: "T101",
					},
				},
			},
			uncompressedPayload: `{
				"config_table": {
					"plugins": [
						{
							"id": "759c0d3a-bc3d-4ccc-8d4d-f92de95c1f1a",
							"name": "plugin_1",
							"config": {
								"plugin_field_1": []
							}
						}
					]
				}
			}`,
			dataPlaneVersion: "3.0.0",
			expectedPayload: `{
				"config_table": {
					"plugins": [
						{
							"id": "759c0d3a-bc3d-4ccc-8d4d-f92de95c1f1a",
							"name": "plugin_1",
							"config": {}
						}
					]
				}
			}`,
			expectedChanges: TrackedChanges{},
		},
```
The second, implicit field removal is not processed. I'm looking for a workaround.

omegabytes: Actually, I just checked the above test case against main and it fails. Does the implicit field removal actually work?
```
Error:      	Not equal: 
        	            	expected: map[string]interface {}{"config_table":map[string]interface {}{"plugins":[]interface {}{map[string]interface {}{"config":map[string]interface {}{}, "id":"759c0d3a-bc3d-4ccc-8d4d-f92de95c1f1a", "name":"plugin_1"}}}}
        	            	actual  : map[string]interface {}{"config_table":map[string]interface {}{"plugins":[]interface {}{map[string]interface {}{"config":map[string]interface {}{"plugin_field_2":[]interface {}{}}, "id":"759c0d3a-bc3d-4ccc-8d4d-f92de95c1f1a", "name":"plugin_1"}}}}
        	            	
        	            	Diff:
        	            	--- Expected
        	            	+++ Actual
        	            	@@ -4,3 +4,5 @@
        	            	    (map[string]interface {}) (len=3) {
        	            	-    (string) (len=6) "config": (map[string]interface {}) {
        	            	+    (string) (len=6) "config": (map[string]interface {}) (len=1) {
        	            	+     (string) (len=14) "plugin_field_2": ([]interface {}) {
        	            	+     }
        	            	     },
```

omegabytes: I'm removing the two failing tests in [ecff883](https://github.com/Kong/koko/pull/469/commits/ecff883f22298c4a60cc2c5fc5e636848d2d7e47).

internal/test/e2e/version_compatibility_test.go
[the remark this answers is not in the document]
omegabytes: I'd need to add a linter override or change the way we read the response body. I'm punting on it for this change as it's already pretty long in the tooth.

internal/server/kong/ws/config/compat/compatibility_table.go
GGabriele: This test fails in main too because you need to at least set one field for the plugin:

```
{
    "code": 3,
    "details": [
        {
            "@type": "type.googleapis.com/kong.admin.model.v1.ErrorDetail",
            "messages": [
                "at least one of these fields must be non-empty: 'config.functions', 'config.certificate', 'config.rewrite', 'config.access', 'config.header_filter', 'config.body_filter', 'config.log'"
            ],
            "type": "ERROR_TYPE_ENTITY"
        }
    ],
    "message": "validation error"
}
```

All good then.

GGabriele (approved):

PR #470, opened 2022-10-14
fix: CA certs validation let certs with CA=False pass

GGabriele:
Right now the CA certs validation is only checking whether the cert's Basic Constraints attribute is valid or not, while it should also check whether the 'CA' basic constraint is set and fail if it's not.

### DIFF
--- a/internal/resource/ca_certificate.go
+++ b/internal/resource/ca_certificate.go
@@ -97,8 +97,8 @@ func (r CACertificate) Validate(ctx context.Context) error {
 			},
 		}
 	}
-	if !cert.BasicConstraintsValid {
-		errStr := `certificate does not appear to be a CA because` +
+	if !cert.BasicConstraintsValid || !cert.IsCA {
+		errStr := `certificate does not appear to be a CA because ` +
 			`it is missing the "CA" basic constraint`
 		return validation.Error{
 			Errs: []*v1.ErrorDetail{
--- a/internal/resource/ca_certificate_test.go
+++ b/internal/resource/ca_certificate_test.go
@@ -76,7 +76,7 @@ p0TGlNicxb2nnhRuh1U63VkBhl3k17vClDiwqHYwe7fPzlmj6kZMaXnpK6rd4kg7
 -----END CERTIFICATE-----
 `
 
-	invalidCert = `
+	invalidCertOne = `
 -----BEGIN CERTIFICATE-----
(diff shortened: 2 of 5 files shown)


GGabriele:
I'd say let's way for [this other PR from Alex](https://github.com/Kong/koko/pull/469) to be merged in before merging this one.

internal/resource/ca_certificate.go
```diff
@@ -97,7 +97,7 @@ func (r CACertificate) Validate(ctx context.Context) error {
 			},
 		}
 	}
-	if !cert.BasicConstraintsValid {
+	if !cert.BasicConstraintsValid || !cert.IsCA {
 		errStr := `certificate does not appear to be a CA because` +
```
tjasko, line 101: Existing bug, but can we ensure there's a space after `because`?

internal/resource/ca_certificate_test.go
```diff
@@ -76,7 +76,7 @@ p0TGlNicxb2nnhRuh1U63VkBhl3k17vClDiwqHYwe7fPzlmj6kZMaXnpK6rd4kg7
 -----END CERTIFICATE-----
 `
 
-	invalidCert = `
+	invalidCertOne = `
```
tjasko, line 79: nit: Should we throw some comments on these certs on how they were created?

GGabriele: It'd be a good idea, if I had notes 🤦

tjasko (approved):

internal/resource/ca_certificate_test.go
tjasko: Yeah I was mostly concerned what it is, allows CA, and whatnot.

internal/resource/ca_certificate.go
```diff
@@ -97,8 +97,8 @@ func (r CACertificate) Validate(ctx context.Context) error {
 			},
 		}
 	}
-	if !cert.BasicConstraintsValid {
-		errStr := `certificate does not appear to be a CA because` +
+	if !cert.BasicConstraintsValid || !cert.IsCA {
+		errStr := `certificate does not appear to be a CA because ` +
```
hbagdi, line 101: Misleading error. If the CA attribute is explicitly set to false then the attribute is present (not missing) but set to false. Can we do better?

GGabriele: Ouch, good catch! What about this? https://github.com/Kong/koko/pull/495

PR #472, opened 2022-10-14
fix: honoring the cluster ID when validating plugins

tjasko:
Corrects an issue where the `POST /v1/schemas/json/plugin/validate` endpoint could not validate custom plugins properly. This also corrects an issue where custom plugin schemas were not pulled in when a cluster ID is not provided.

This is in draft status as we should update the tests as well.

### DIFF
--- a/internal/plugin/validators/lua_validator.go
+++ b/internal/plugin/validators/lua_validator.go
@@ -354,7 +354,7 @@ func (v *LuaValidator) getDB(ctx context.Context) (store.Store, error) {
 	}
 	reqCluster, ok := ctx.Value(util.ContextKeyCluster).(*grpcModel.RequestCluster)
 	if !ok {
-		reqCluster = &grpcModel.RequestCluster{Id: "default"}
+		reqCluster = &grpcModel.RequestCluster{Id: store.DefaultCluster}
 	}
 
 	store, err := v.storeLoader.Load(ctx, reqCluster)
--- a/internal/plugin/validators/lua_validator.go
+++ b/internal/plugin/validators/lua_validator.go
@@ -354,7 +354,7 @@ func (v *LuaValidator) getDB(ctx context.Context) (store.Store, error) {
 	}
 	reqCluster, ok := ctx.Value(util.ContextKeyCluster).(*grpcModel.RequestCluster)
 	if !ok {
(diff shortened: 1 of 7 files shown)


internal/plugin/validators/lua_validator_test.go
```diff
@@ -1645,10 +1645,10 @@ func TestLuaValidator_GetDB(t *testing.T) {
 	storeLoader := setupStoreLoader(t)
 	require.NotNil(t, storeLoader)
 	goodValidator.storeLoader = storeLoader
-	t.Run("context not containing RequestCluster returns error", func(t *testing.T) {
+	t.Run("context not containing RequestCluster returns default store", func(t *testing.T) {
 		db, err := goodValidator.getDB(context.Background())
-		assert.Nil(t, db)
-		assert.Error(t, err, "invalid context: failed to retrieve RequestCluster from context")
+		require.NoError(t, err)
+		assert.NotNil(t, db)
```
tjasko, line 1651: We should likely track some work to expose a `Cluster() string` method on the DB, so we can actually assert this better in tests... as well, no other way to check without the use of nasty reflection.

internal/plugin/validators/lua_validator.go
```diff
@@ -354,7 +354,7 @@ func (v *LuaValidator) getDB(ctx context.Context) (store.Store, error) {
 	}
 	reqCluster, ok := ctx.Value(util.ContextKeyCluster).(*grpcModel.RequestCluster)
 	if !ok {
-		return nil, errors.New("invalid context: failed to retrieve RequestCluster from context")
+		reqCluster = &grpcModel.RequestCluster{Id: "default"}
```
tjasko, line 357: We really should store `default` as a constant in the codebase & use that across everything. Don't want to make that change in this PR, though perhaps we should track some work for the change.

internal/plugin/validators/lua_validator_test.go
tjasko: Implemented in Kong/koko#482.

internal/plugin/validators/lua_validator.go
tjasko: Implemented in https://github.com/Kong/koko/pull/482.

GGabriele (approved):

PR #483, opened 2022-10-18
feat: remove core entities for compatibility

EJKinger:
This adds a function to remove core entities when required by the
compatibility table and to track those changes.

### DIFF
--- a/internal/server/kong/ws/config/version_compatibility.go
+++ b/internal/server/kong/ws/config/version_compatibility.go
@@ -263,6 +263,42 @@ func (vc *WSVersionCompatibility) processConfigTableUpdates(uncompressedPayload 
 	return processedPayload, nil
 }
 
+func (vc *WSVersionCompatibility) removeCoreEntity(
+	processedPayload string,
+	entityType string,
+	dataPlaneVersionStr string,
+	configTableKey string,
+	changeID ChangeID,
+	tracker *ChangeTracker,
+) string {
+	entityIDs := gjson.Get(processedPayload, fmt.Sprintf("config_table.%s.#.id", configTableKey))
+	for _, id := range entityIDs.Array() {
+		err := tracker.TrackForResource(changeID, ResourceInfo{
+			Type: entityType,
+			ID:   id.String(),
+		})
+		if err != nil {
+			vc.logger.Error("failed to track version compatibility change",
+				zap.String("change-id", string(changeID)),
+				zap.String("resource-type", entityType))
+		}
+	}
+
+	processedPayload, err := sjson.Delete(processedPayload, fmt.Sprintf("config_table.%s", configTableKey))
+	if err != nil {
+		vc.logger.With(zap.String("entity", entityType)).
+			With(zap.String("data-plane", dataPlaneVersionStr)).
+			With(zap.Error(err)).
+			Error("entity was not removed from configuration")
+	} else {
+		vc.logger.With(zap.String("entity", entityType)).
+			With(zap.String("data-plane", dataPlaneVersionStr)).
+			Warn("removing entity which is incompatible with data plane")
+	}
+
+	return processedPayload
+}
+
 func (vc *WSVersionCompatibility) removePlugin(
 	processedPayload string,
 	pluginName string,
@@ -694,6 +730,10 @@ func (vc *WSVersionCompatibility) processCoreEntityUpdates(payload string,
 			With(zap.String("data-plane", dataPlaneVersionStr)).
 			With(zap.Error(err)).
 			Error("error while updating entities")
+	}
+	if configTableUpdate.Remove && results.Exists() {
+		processedPayload = vc.removeCoreEntity(processedPayload, entityType,
+			dataPlaneVersionStr, configTableKey, configTableUpdate.ChangeID, tracker)
 	}
 	return processedPayload
 }
--- a/internal/server/kong/ws/config/version_compatibility_test.go
+++ b/internal/server/kong/ws/config/version_compatibility_test.go
@@ -6070,6 +6070,210 @@ func TestVersionCompatibility_ProcessConfigTableUpdates(t *testing.T) {
 			}`,
 			expectedChanges: TrackedChanges{},
 		},
+		{
+			name: "remove core entities when indicated",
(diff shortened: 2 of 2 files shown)


EJKinger:
Testing will be added as a part of #464, as vaults will be the first entity to actually make use of this feature.

internal/server/kong/ws/config/version_compatibility.go
```diff
@@ -263,6 +263,44 @@ func (vc *WSVersionCompatibility) processConfigTableUpdates(uncompressedPayload
...
 
+func (vc *WSVersionCompatibility) removeCoreEntity(
+	processedPayload string,
+	entityType string,
+	dataPlaneVersionStr string,
+	configTableKey string,
+	changeID ChangeID,
+	tracker *ChangeTracker,
+) string {
+	entity := gjson.Get(processedPayload, fmt.Sprintf("config_table.%ss", entityType))
```
mikefero, line 274:
```suggestion
	entity := gjson.Get(processedPayload, fmt.Sprintf("config_table.%s", entityType))
```

internal/server/kong/ws/config/version_compatibility.go
```diff
@@ -263,6 +263,44 @@ func (vc *WSVersionCompatibility) processConfigTableUpdates(uncompressedPayload
...
+			ID:   id,
+		})
+		if err != nil {
+			vc.logger.Error("failed to track version compatibility change",
+				zap.String("change-id", string(changeID)),
+				zap.String("resource-type", entityType))
+		}
+	}
+
+	processedPayload, err := sjson.Delete(processedPayload, fmt.Sprintf("config_table.%ss", configTableKey))
```
mikefero, line 289:
```suggestion
	processedPayload, err := sjson.Delete(processedPayload, fmt.Sprintf("config_table.%s", configTableKey))
```

mikefero (changes requested):
> Testing will be added as a part of https://github.com/Kong/koko/pull/464, as vaults will be the first entity to actually make use of this feature.

Do the unit tests exist in #464 for the `Remove`? Maybe we should move them here as I found an issue with `gjson` lookup for `entity`. WDYT?

internal/server/kong/ws/config/version_compatibility.go
EJKinger: The testing is in #464 using entity deletion rules for the Vaults.  I'd rather do the testing in this PR, but it felt silly to write mock entity deletion rules for an entity that we will never want to delete.

internal/server/kong/ws/config/version_compatibility.go
```diff
@@ -263,6 +263,44 @@ func (vc *WSVersionCompatibility) processConfigTableUpdates(uncompressedPayload
...
+	dataPlaneVersionStr string,
+	configTableKey string,
+	changeID ChangeID,
+	tracker *ChangeTracker,
+) string {
+	entity := gjson.Get(processedPayload, fmt.Sprintf("config_table.%s", configTableKey))
+	for _, res := range entity.Array() {
+		var err error
+		id := res.Get("id").String()
+		err = tracker.TrackForResource(changeID, ResourceInfo{
```
tjasko, line 278: nit: Can get rid of `var err error` and just do `err := tracker...` instead

internal/server/kong/ws/config/version_compatibility.go
```diff
@@ -263,6 +263,44 @@ func (vc *WSVersionCompatibility) processConfigTableUpdates(uncompressedPayload
...
 
+func (vc *WSVersionCompatibility) removeCoreEntity(
+	processedPayload string,
+	entityType string,
+	dataPlaneVersionStr string,
+	configTableKey string,
+	changeID ChangeID,
+	tracker *ChangeTracker,
+) string {
+	entity := gjson.Get(processedPayload, fmt.Sprintf("config_table.%s", configTableKey))
```
tjasko, line 274: I would just fetch the ID as that's all we need.
```go
entityIDs := gjson.Get(processedPayload, fmt.Sprintf("config_table.%s.#.id", configTableKey))
```

internal/server/kong/ws/config/version_compatibility.go
EJKinger: fixed this.

internal/server/kong/ws/config/version_compatibility.go
EJKinger: fixed this.

internal/server/kong/ws/config/version_compatibility.go
EJKinger: fixed this.

internal/server/kong/ws/config/version_compatibility.go
EJKinger: I moved the testing into this PR.

EJKinger:
> > Testing will be added as a part of #464, as vaults will be the first entity to actually make use of this feature.
> 
> Do the unit tests exist in #464 for the `Remove`? Maybe we should move them here as I found an issue with `gjson` lookup for `entity`. WDYT?

Testing has been moved into this PR.

GGabriele (approved):

tjasko (approved):
Great work 😄

internal/server/kong/ws/config/version_compatibility.go
```diff
@@ -263,6 +263,42 @@ func (vc *WSVersionCompatibility) processConfigTableUpdates(uncompressedPayload
...
+	processedPayload string,
+	entityType string,
+	dataPlaneVersionStr string,
+	configTableKey string,
+	changeID ChangeID,
+	tracker *ChangeTracker,
+) string {
+	entityIDs := gjson.Get(processedPayload, fmt.Sprintf("config_table.%s.#.id", configTableKey))
+	for _, id := range entityIDs.Array() {
+		err := tracker.TrackForResource(changeID, ResourceInfo{
```
hbagdi, line 276: Please update this to use Track() instead. When a change affects *all* resources, global tracking must be used. This is to ensure that the user fixes the problem at a global level rather than trying to figure out the problem within the context of a resource (which could mislead the user).

PR #484, opened 2022-10-21
fix: authenticators should always return ErrAuth

GGabriele:


aboudreault (approved):

tjasko (approved):
The tests for this could definitely be improved, as the coverage is a bit weak & they're not asserting the actual error (only that one exists), but it's unrelated to this change anyway.

I would recommend for us to fix that up, however, I'll accept this PR anyway as it's an existing issue.

PR #485, opened 2022-10-21
feat: add dataplane metrics for control plane

omegabytes:
Adds the following metrics:
    * duration (s) to establish initial connection between dp and cp
    * duration (s) of individual config mutations
    * duration (s) of total aggregate config mutations
    * duration (s) of version compat cache miss path

### DIFF
--- a/internal/server/kong/ws/config/configuration.go
+++ b/internal/server/kong/ws/config/configuration.go
@@ -57,10 +57,9 @@ func (l *KongConfigurationLoader) Load(ctx context.Context, clusterID string) (C
 		err := m.Mutate(ctx, MutatorOpts{ClusterID: clusterID},
 			configTable)
 		metrics.Histogram(
-			"data_plane_config_mutation_individual_duration_seconds",
+			"config_mutation_individual_duration_seconds",
 			time.Since(mutationStartTime).Seconds(),
 			metrics.Tag{Key: "status", Value: lo.Ternary(err == nil, "success", "fail")},
-			metrics.Tag{Key: "protocol", Value: "ws"},
 			metrics.Tag{Key: "mutator_name", Value: m.Name()},
 		)
 		if err != nil {
--- a/internal/server/kong/ws/config/configuration.go
+++ b/internal/server/kong/ws/config/configuration.go
@@ -9,18 +9,8 @@ import (
 	"time"
 
 	"github.com/kong/koko/internal/json"
-	metricsv2 "github.com/kong/koko/internal/metrics/v2"
-	"github.com/prometheus/client_golang/prometheus"
+	"github.com/kong/koko/internal/metrics"
 )
-
-var singleMutationDuration = metricsv2.NewHistogram(
-	prometheus.NewRegistry(),
-	metricsv2.HistogramOpts{
-		Subsystem:  "cp",
-		Name:       "data_plane_config_mutation_time_individual",
-		Help:       "Duration in ms it takes an individual mutator to reconfigure a dataplane payload",
-		LabelNames: []string{"status", "mutator_name"},
-	})
 
 type Map map[string]interface{}
 
@@ -67,23 +57,25 @@ func (l *KongConfigurationLoader) Load(ctx context.Context, clusterID string) (C
 			configTable)
 		mutationDuration := time.Since(mutationStartTime).Milliseconds()
 		if err != nil {
-			singleMutationDuration.Observe(float64(mutationDuration),
-				metricsv2.Label{
+			metrics.Histogram("data_plane_config_mutation_time_individual",
+				float64(mutationDuration),
+				metrics.Tag{
 					Key:   "status",
 					Value: "fail",
 				},
-				metricsv2.Label{
+				metrics.Tag{
 					Key:   "mutator_name",
 					Value: m.Name(),
 				})
 			return Content{}, err
 		}
-		singleMutationDuration.Observe(float64(mutationDuration),
-			metricsv2.Label{
+		metrics.Histogram("data_plane_config_mutation_time_individual",
+			float64(mutationDuration),
+			metrics.Tag{
 				Key:   "status",
 				Value: "success",
 			},
-			metricsv2.Label{
+			metrics.Tag{
 				Key:   "mutator_name",
 				Value: m.Name(),
 			})
--- a/internal/server/kong/ws/config/configuration.go
+++ b/internal/server/kong/ws/config/configuration.go
@@ -6,9 +6,21 @@ import (
 	"context"
 	"fmt"
 	"io"
+	"time"
 
 	"github.com/kong/koko/internal/json"
+	metricsv2 "github.com/kong/koko/internal/metrics/v2"
+	"github.com/prometheus/client_golang/prometheus"
 )
+
+var singleMutationDuration = metricsv2.NewHistogram(
+	prometheus.NewRegistry(),
+	metricsv2.HistogramOpts{
+		Subsystem:  "cp",
+		Name:       "data_plane_config_mutation_time_individual",
+		Help:       "Duration in ms it takes an individual mutator to reconfigure a dataplane payload",
+		LabelNames: []string{"status", "mutator_name"},
+	})
 
 type Map map[string]interface{}
 
@@ -50,11 +62,31 @@ func (l *KongConfigurationLoader) Register(mutator Mutator) error {
 func (l *KongConfigurationLoader) Load(ctx context.Context, clusterID string) (Content, error) {
 	var configTable DataPlaneConfig = map[string]interface{}{}
 	for _, m := range l.mutators {
+		mutationStartTime := time.Now()
 		err := m.Mutate(ctx, MutatorOpts{ClusterID: clusterID},
 			configTable)
+		mutationDuration := time.Since(mutationStartTime).Milliseconds()
 		if err != nil {
+			singleMutationDuration.Observe(float64(mutationDuration),
+				metricsv2.Label{
+					Key:   "status",
+					Value: "fail",
+				},
+				metricsv2.Label{
+					Key:   "mutator_name",
+					Value: m.Name(),
+				})
 			return Content{}, err
 		}
+		singleMutationDuration.Observe(float64(mutationDuration),
+			metricsv2.Label{
+				Key:   "status",
+				Value: "success",
+			},
+			metricsv2.Label{
+				Key:   "mutator_name",
+				Value: m.Name(),
+			})
 	}
 
 	return ReconfigurePayload(configTable)
--- a/internal/server/kong/ws/config/configuration.go
+++ b/internal/server/kong/ws/config/configuration.go
@@ -10,6 +10,7 @@ import (
 
 	"github.com/kong/koko/internal/json"
 	"github.com/kong/koko/internal/metrics"
+	"github.com/samber/lo"
 )
 
 type Map map[string]interface{}
@@ -55,30 +56,16 @@ func (l *KongConfigurationLoader) Load(ctx context.Context, clusterID string) (C
 		mutationStartTime := time.Now()
 		err := m.Mutate(ctx, MutatorOpts{ClusterID: clusterID},
 			configTable)
-		mutationDuration := time.Since(mutationStartTime).Milliseconds()
+		metrics.Histogram(
+			"data_plane_config_mutation_individual_duration_seconds",
+			time.Since(mutationStartTime).Seconds(),
+			metrics.Tag{Key: "status", Value: lo.Ternary(err == nil, "success", "fail")},
+			metrics.Tag{Key: "protocol", Value: "ws"},
+			metrics.Tag{Key: "mutator_name", Value: m.Name()},
+		)
 		if err != nil {
-			metrics.Histogram("data_plane_config_mutation_time_individual",
-				float64(mutationDuration),
-				metrics.Tag{
-					Key:   "status",
-					Value: "fail",
-				},
-				metrics.Tag{
-					Key:   "mutator_name",
-					Value: m.Name(),
-				})
 			return Content{}, err
 		}
-		metrics.Histogram("data_plane_config_mutation_time_individual",
-			float64(mutationDuration),
-			metrics.Tag{
-				Key:   "status",
-				Value: "success",
-			},
-			metrics.Tag{
-				Key:   "mutator_name",
-				Value: m.Name(),
-			})
 	}
 
 	return ReconfigurePayload(configTable)
--- a/internal/server/kong/ws/manager.go
+++ b/internal/server/kong/ws/manager.go
@@ -25,6 +25,7 @@ import (
 	"github.com/kong/koko/internal/resource"
 	"github.com/kong/koko/internal/server/kong/ws/config"
 	"github.com/kong/koko/internal/store"
+	"github.com/samber/lo"
 	"go.uber.org/atomic"
 	"go.uber.org/zap"
 	"google.golang.org/grpc/codes"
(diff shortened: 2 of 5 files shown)


internal/server/kong/ws/config/configuration.go
```diff
@@ -50,11 +52,33 @@ func (l *KongConfigurationLoader) Register(mutator Mutator) error {
...
+		metrics.Histogram("data_plane_config_mutation_time_individual",
+			float64(mutationDuration),
+			metrics.Tag{
+				Key:   "status",
+				Value: "success",
+			},
+			metrics.Tag{
+				Key:   "mutator_name",
+				Value: m.Name(),
+			})
```
tjasko, line 81: This (and other instances of this) can be simplified: 😃 
```suggestion
		mutationStartTime := time.Now()
		err := m.Mutate(ctx, MutatorOpts{ClusterID: clusterID},
			configTable)
		metrics.Histogram(
			"data_plane_config_mutation_time_individual",
			float64(time.Since(mutationStartTime).Milliseconds()),
			metrics.Tag{Key: "status", Value: lo.Ternary(err == nil, "success", "fail")},
			metrics.Tag{Key: "mutator_name", Value: m.Name()},
		)
		if err != nil {
			return Content{}, err
		}
```

internal/server/kong/ws/config/configuration.go
```diff
@@ -50,11 +52,33 @@ func (l *KongConfigurationLoader) Register(mutator Mutator) error {
 func (l *KongConfigurationLoader) Load(ctx context.Context, clusterID string) (Content, error) {
 	var configTable DataPlaneConfig = map[string]interface{}{}
 	for _, m := range l.mutators {
+		mutationStartTime := time.Now()
 		err := m.Mutate(ctx, MutatorOpts{ClusterID: clusterID},
 			configTable)
+		mutationDuration := time.Since(mutationStartTime).Milliseconds()
 		if err != nil {
+			metrics.Histogram("data_plane_config_mutation_time_individual",
```
tjasko, line 60: Can we follow standard Prometheus conventions: 

1. Naming this metric series: `data_plane_config_mutation_individual_duration_seconds`
    - We're not emitting the entire mutation process as a metric. Do we want to be doing that too?
2. Altering to push seconds instead (with a decimals)

internal/server/kong/ws/server.go
```diff
@@ -119,6 +122,13 @@ func (h websocketHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 		loggerWithNode.Error("failed to upgrade websocket connection", zap.Error(err))
 		return
 	}
+	dpConnectionDuration := time.Since(dpConnectStartTime).Milliseconds()
+	metrics.Histogram("data_plane_time_to_connect",
```
tjasko, line 126: Related to the other comments, can we use seconds here & rename to `data_plane_connect_duration_seconds`.

Also, should we rename this to `data_plane_websocket_connect_duration_seconds` to be a bit more specific?

internal/server/kong/ws/server.go
```diff
@@ -85,6 +87,7 @@ func validateRequest(r *http.Request) error {
 }
 
 func (h websocketHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
+	dpConnectStartTime := time.Now()
```
tjasko, line 90: Should we move this right above `m, err := h.authenticator.Authenticate(r)`? As the way it currently is, it'd capture a bit more than just connecting.

internal/server/kong/ws/node.go
```diff
@@ -185,6 +187,12 @@ func (n *Node) readThread() error {
 
 func (n *Node) sendConfig(ctx context.Context, payload *Payload) error {
 	content, err := n.getPayload(ctx, payload)
+	metrics.Histogram("data_plane_config_size",
+		float64(unsafe.Sizeof(content)),
```
tjasko, line 191: This won't do what you expect, as it's grabbing the size of the entire `config.Content` object & not the size of the actual config.

Let's be sure we're capturing the size of the JSON payload being sent to the data plane, and ideally, it's uncompressed size as that's what matters. Let's modify the `config.Content` struct to include the size in bytes as well, so that we can log it here.

tjasko (changes requested):

internal/server/kong/ws/config/configuration.go
omegabytes: 1. [This is tracks total](https://github.com/Kong/koko/pull/485/files#diff-4b5bf8ad94aaa24b01cab08505b8ac550fc4efe48eac16233df472395212c06dR481), which accounts for success and fail states as well
2. I take this to mean `t=.001s` vs `t=1ms`. I'm accustomed to emitting ms, is emitting decimal seconds a prometheus convention?

internal/server/kong/ws/server.go
omegabytes: @mikefero wdyt? You mentioned we should include request validation when we talked about this metric. I don't have a strong opinion either way.

internal/server/kong/ws/config/configuration.go
tjasko: 1. Yes, that's fine.
2. Derp, I forgot to link the doc. Yes, it's a convention. https://prometheus.io/docs/practices/naming/

internal/server/kong/ws/server.go
tjasko: The strange thing to me is the naming, as the metric is named to make it think it's the connection duration only. It's fine for including the request validation, but it seemed a little strange.

Either way, both are part of establishing a connection. Whenever we switch this over to the Prometheus metric format, we can at least provide a help message documenting what it's capturing.

mikefero: I am fine with having the name indicate this is a validated authenticated connection duration; more verbosity in the name doesn't hurt as we could add the other metrics during the connection phases later on.

internal/server/kong/ws/server.go
omegabytes: I've seen us use a "protocol": "ws" metric tag in other places. I don't have an opinion either way.

internal/server/kong/ws/node.go
omegabytes: I will address this as a separate issue.

omegabytes:
> Should we use a `cp_` prefix for these metrics?

Are you suggesting something like `cp_data_plane_connect_duration_seconds`?

GGabriele:
> > Should we use a `cp_` prefix for these metrics?
> 
> Are you suggesting something like `cp_data_plane_connect_duration_seconds`?

Yes, basically indicating which koko component is generating these metrics.

internal/server/kong/ws/server.go
tjasko: > I've seen us use a "protocol": "ws" metric tag in other places. I don't have an opinion either way.
At least right now, there's only one protocol here, as it's all websockets anyway (for both wRPC & the older WS protocol). We don't need to rename it to be a bit more specific, was mostly just a question whether we should.

internal/server/kong/ws/server.go
tjasko: Ohh I'm fine with how it is right now, just that it's slightly confusing as mentioned previously. Technically speaking this is the time for authentication & the actual connection. Perhaps instead of saying "connect", "authenticate" may be a better word.

internal/server/kong/ws/node.go
tjasko: That works with me, thanks for removing this metric then.

internal/server/kong/ws/payload.go
```diff
@@ -136,6 +138,12 @@ func (p *Payload) configForVersion(version string) (cacheEntry, error) {
 		}
 		// cache it
 		err = p.configCache.store(version, entry)
+		configUpdateProcessingDuration := time.Since(configUpdateProcessingStartTime).Seconds()
+		metrics.Histogram("data_plane_version_compatibility_duration_seconds", configUpdateProcessingDuration,
+			metrics.Tag{Key: "dp_version", Value: version},
+			metrics.Tag{Key: "status", Value: "success"},
```
tjasko, line 144: Did we want to update the status here to reflect whether `p.vc.ProcessConfigTableUpdates()` errored?

Additionally, this is capturing the duration of the config loading + storing as well, so it's technically more than just version compatibility, though at least right now, the majority of this code is focusing on versions compatibility right now. Perhaps well want to move this metric emitting code below the error handling logic for `p.vc.ProcessConfigTableUpdates()` & additionally move the `configUpdateProcessingStartTime` assignment right before the call. This is bit of a nit, as it shouldn't skew the metric results too badly,  but it could since this part of the code is blocked by a mutex.

omegabytes: Excluded caching from processing metric. We can look at instrumenting cache hit/miss/set/error behavior in the future.
[bb17fd1](https://github.com/Kong/koko/pull/485/commits/bb17fd17013470fbc21afa2117546948ca2e9d0d)

tjasko: Awesome, looks good with me! Thanks for moving the logic around, I was mostly concerned about the mutex throwing off the validity of this metric, so appreciate that change!

tjasko (dismissed):

internal/server/kong/ws/config/configuration.go
```diff
@@ -6,8 +6,11 @@ import (
 	"context"
```
hbagdi, line 6: General comment: Are we sure we want to track at a seconds granularity? Isn't that too large?

internal/server/kong/ws/server.go
```diff
@@ -119,6 +122,12 @@ func (h websocketHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 		loggerWithNode.Error("failed to upgrade websocket connection", zap.Error(err))
 		return
 	}
+	// Tracks how long it takes to establish the connection between the control plane and the data plane.
+	metrics.Histogram("data_plane_connect_duration_seconds",
```
hbagdi, line 126:
```suggestion
	metrics.Histogram("wrpc_connection_setup_duration_seconds",
```

Once again, do not prefix with DP or CP, those qualifiers are not necessary.

internal/server/kong/ws/config/configuration.go
```diff
@@ -50,8 +53,16 @@ func (l *KongConfigurationLoader) Register(mutator Mutator) error {
 func (l *KongConfigurationLoader) Load(ctx context.Context, clusterID string) (Content, error) {
 	var configTable DataPlaneConfig = map[string]interface{}{}
 	for _, m := range l.mutators {
+		mutationStartTime := time.Now()
 		err := m.Mutate(ctx, MutatorOpts{ClusterID: clusterID},
 			configTable)
+		metrics.Histogram(
+			"data_plane_config_mutation_individual_duration_seconds",
+			time.Since(mutationStartTime).Seconds(),
+			metrics.Tag{Key: "status", Value: lo.Ternary(err == nil, "success", "fail")},
```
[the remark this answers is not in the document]
omegabytes: Histogram of `data_plane_config_mutation_individual_duration_seconds` represents count/sum/quantiles etc of all observations for this individual mutations. In my opinion, success and fail cases are a subtype of the metric and should be handled with tags. This is the same reason I use a mutator tag with the mutator name instead of appending the mutator name to the metric name.

To extend the reasoning a bit: imagine we have `config_mutation_individual_duration_seconds_fail` and want to emit metrics about why the failure happened. In this case, I'd include a `reason` tag with potential (example) values such as "internal" , "bad_input", "invalid_type", "timeout" etc. Or, I could append that reason to the metric and end up with 4 metric names for one very similar data point.

Extended further, if we append a reason to all our metrics then we are creating a ton of metric names. We also need to choose if we want to instrument those separate metrics in many places, or interpolate reason into the metric name to save on duplication of non-feature code.

internal/server/kong/ws/manager.go
```diff
@@ -474,7 +475,14 @@ type ConfigClient struct {
 }
 
 func (m *Manager) reconcileKongPayload(ctx context.Context) error {
+	mutationStartTime := time.Now()
 	config, err := m.configLoader.Load(ctx, m.Cluster.Get())
+	metrics.Histogram(
+		"data_plane_config_mutation_total_duration_seconds",
+		time.Since(mutationStartTime).Seconds(),
+		metrics.Tag{Key: "status", Value: lo.Ternary(err == nil, "success", "fail")},
+		metrics.Tag{Key: "protocol", Value: "ws"},
```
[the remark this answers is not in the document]
omegabytes: This is an aggregate of https://github.com/Kong/koko/pull/485/files#diff-e0a3738e9a2e7a46592b82fb6819a25aa35998cefcec4dfe2e22a9b99cd0f48fR60, including `ReconfigurePayload` execution time. If we do not want to include `ReconfigurePayload` execution time, we can remove this and instead take the sum of `data_plane_config_mutation_individual_duration_seconds`

internal/server/kong/ws/config/configuration.go
omegabytes: I agree - I am accustomed to ms as the base unit. However, @tjasko pointed out that Prometheus convention is to use seconds as the base unit: https://github.com/Kong/koko/pull/485#discussion_r1003676099

omegabytes:
@hbagdi I've made the requested changes. Note that [I am following established convention in the code](https://github.com/Kong/koko/blob/4809bbf758d1bb92e4ed5290310d0b65e1341a43/internal/server/kong/ws/config_service.go#L67-L75) where possible, and review comments indicate we are no longer following that pattern.

internal/server/kong/ws/config/configuration.go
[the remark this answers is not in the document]
tjasko: The Prometheus [convention](https://prometheus.io/docs/practices/naming/) states:
> ...should represent the same logical thing-being-measured across all label dimensions.

Because we're logging config mutation duration here, regardless of if it passed/failed, we're still logging how long it took. Also because we're not conditionally adding any labels based on whether it passed/failed, this particular metric is okay IMHO. In the event we wanted to log more labels that was specific to only one edge-case (e.g.: when it fails), then I'd say it should be split out into its own metric.

This is honestly up for interpretation though, I've seen it done both ways, but that's my 2¢.

internal/server/kong/ws/config/configuration.go
```diff
@@ -50,8 +53,16 @@ func (l *KongConfigurationLoader) Register(mutator Mutator) error {
...
+			metrics.Tag{Key: "protocol", Value: "ws"},
```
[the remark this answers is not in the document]
tjasko: I did comment on something similar in a chat with Alex, and I do agree here, as the protocol doesn't impact this area of the code.

internal/server/kong/ws/config/configuration.go
```diff
@@ -50,8 +53,16 @@ func (l *KongConfigurationLoader) Register(mutator Mutator) error {
 func (l *KongConfigurationLoader) Load(ctx context.Context, clusterID string) (Content, error) {
 	var configTable DataPlaneConfig = map[string]interface{}{}
 	for _, m := range l.mutators {
+		mutationStartTime := time.Now()
 		err := m.Mutate(ctx, MutatorOpts{ClusterID: clusterID},
 			configTable)
+		metrics.Histogram(
+			"data_plane_config_mutation_individual_duration_seconds",
```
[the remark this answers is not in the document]
tjasko: We definitely do need to figure out a better standard for namespacing our metrics, as right now, it's not very well thought of. I've already brought this up with Alex regarding the metrics v2 implementation.

I could go either way here, the configuration is being pushed to the data plane, and keeping `data_plane` in it signals that it's the config being pushed to the DP, and not some other config. How can we make this metric name obvious in what it does when someone is reading all the metrics being emitted by Koko? How this metric is currently named helps with that.

internal/server/kong/ws/payload.go
```diff
@@ -126,6 +129,12 @@ func (p *Payload) configForVersion(version string) (cacheEntry, error) {
 					Value: version,
 				})
 		}
+		configUpdateProcessingDuration := time.Since(configUpdateProcessingStartTime).Seconds()
+		metrics.Histogram("data_plane_version_compatibility_duration_seconds", configUpdateProcessingDuration,
```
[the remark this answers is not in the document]
tjasko: Given that we already have metrics called `version_compatibility_configuration_change_count`, perhaps we shouldn't deviate from the existing standards and just call it `version_compatibility_configuration_duration_seconds`.

As mentioned [previously](https://github.com/Kong/koko/pull/485#discussion_r1008457397), we do need to introduce some proper metric series namespacing, but we can get things cleaned up in the v2 integration & take it more seriously.

internal/server/kong/ws/server.go
tjasko: This code isn't just limited to wRPC, and given that we have an existing metric named `data_plane_ping_total`, this is following current conventions, right?

mikefero: It makes sense to keep these void of qualifiers. I know @omegabytes you have already committed the change to `wrpc_connection_setup_duration_seconds` but as @tjasko pointed out it does make sense to not include that qualifier either.

`connection_setup_duration_seconds` seems appropriate.

mikefero (approved):
One small comment I left around `wprc_connection_setup_duration_seconds`, but other than that this looks good to go.

@tjasko mentioned [this](https://github.com/Kong/koko/pull/485#discussion_r1008459160) but I think `configuration_transformation_duration_seconds` is acceptable and when v2 comes around we can address the `version_compatibility` namespacing.

Please get a +1 from @tjasko; all other items have been addressed.

tjasko (approved):

PR #490, opened 2022-10-24
Fix/sync entity schemas

GGabriele:
Right now JSON entity schemas are missing most default values. Let's just add them so that these defaults can be exposed in the schemas endpoints

### DIFF
--- a/internal/model/json/validation/typedefs/fields.go
+++ b/internal/model/json/validation/typedefs/fields.go
@@ -16,6 +16,8 @@ const (
 	maxPathLength     = 1024
 	hostnamePattern   = "[a-zA-Z0-9]([-a-zA-Z0-9]*[a-zA-Z0-9])?(.[a-zA-Z0-9]([-a-zA-Z0-9]*[a-zA-Z0-9])?)*$"
 
+	DefaultTimeout = 60000
+
 	// Allow alphanumerics, dashes, underscores, colons, periods, and tildes anywhere
 	// within the tag. Spaces are allowed, but not as the first/last character.
 	//
@@ -59,6 +61,7 @@ var Timeout = &generator.Schema{
 	Type:    "integer",
 	Minimum: intP(1),
 	Maximum: maxTimeout,
+	Default: DefaultTimeout,
 }
 
 var Protocol = &generator.Schema{
--- a/internal/resource/service.go
+++ b/internal/resource/service.go
@@ -3,6 +3,7 @@ package resource
 import (
 	"context"
 	"fmt"
+	"math"
 
 	"github.com/imdario/mergo"
 	v1 "github.com/kong/koko/internal/gen/grpc/kong/admin/model/v1"
@@ -15,11 +16,12 @@ import (
 )
 
 const (
-	defaultTimeout = 60000
-	defaultRetries = 5
-	defaultPort    = 80
-	maxRetries     = 32767
-	maxVerifyDepth = 64
+	defaultRetries  = 5
+	defaultPort     = 80
+	maxPort         = math.MaxUint16
+	maxRetries      = math.MaxInt16
+	maxVerifyDepth  = 64
+	defaultProtocol = typedefs.ProtocolHTTP
 	// TypeService denotes the Service type.
 	TypeService = model.Type("service")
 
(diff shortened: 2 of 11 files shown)


internal/resource/service.go
```diff
@@ -15,11 +15,12 @@ import (
...
 const (
-	defaultTimeout = 60000
-	defaultRetries = 5
-	defaultPort    = 80
-	maxRetries     = 32767
-	maxVerifyDepth = 64
+	defaultRetries  = 5
+	defaultPort     = 80
+	maxPort         = 65535
+	maxRetries      = 32767
```
tjasko, line 21: nit:
```suggestion
	maxRetries      = math.MaxInt16
```

internal/resource/service.go
```diff
@@ -15,11 +15,12 @@ import (
...
 
 const (
-	defaultTimeout = 60000
-	defaultRetries = 5
-	defaultPort    = 80
-	maxRetries     = 32767
-	maxVerifyDepth = 64
+	defaultRetries  = 5
+	defaultPort     = 80
+	maxPort         = 65535
```
tjasko, line 20: nit:
```suggestion
	maxPort         = math.MaxUint16
```

internal/resource/upstream.go
```diff
@@ -56,30 +67,68 @@ var (
...
+		"tcp_failures":  0,
+		"http_statuses": defaultPassiveUnhealtyHTTPStatuses,
+		"timeouts":      0,
+	}
+	defaultActive = map[string]interface{}{
+		"concurrency":              defaultConcurrency,
+		"healthy":                  defaultActiveHealthy,
+		"http_path":                "/",
+		"https_verify_certificate": true,
+		"type":                     "http",
```
tjasko, line 118: nit: 
```suggestion
		"type":                     typedefs.ProtocolHTTP,
```

internal/resource/upstream.go
```diff
@@ -56,30 +67,68 @@ var (
...
+		"timeout":                  1,
+		"unhealthy":                defaultActiveUnhealthy,
+	}
+	defaultPassive = map[string]interface{}{
+		"healthy":   defaultPassiveHealthy,
+		"type":      "http",
```
tjasko, line 124: nit:
```suggestion
		"type":      typedefs.ProtocolHTTP,
```

internal/resource/upstream.go
```diff
@@ -122,6 +171,27 @@ var (
 			"grpc",
 			"grpcs",
 		},
+		Default: "http",
```
tjasko, line 174: nit: 
```suggestion
		Default: typedefs.ProtocolHTTP,
```

internal/model/json/validation/typedefs/fields.go
```diff
@@ -16,6 +16,8 @@ const (
 	maxPathLength     = 1024
 	hostnamePattern   = "[a-zA-Z0-9]([-a-zA-Z0-9]*[a-zA-Z0-9])?(.[a-zA-Z0-9]([-a-zA-Z0-9]*[a-zA-Z0-9])?)*$"
```
tjasko, line 18: Should we write a unit test that creates each resource & asserts the non-zero values that were set to the relevant JSON schema? This way, we could have the test fail in the event a default value is missing on the JSON schema.

GGabriele: Right now we already have something like this checking if a resource with minimal configuration matches the default value: https://github.com/Kong/koko/blob/main/internal/resource/route_test.go#L38-L47

Do you think that's sufficient?

tjasko: I was actually talking about asserting that we're not missing any defaults on the JSON schema definition. What you're speaking of is a bit different & is simply just testing the defaults from the Kong Lua code are applied, not that we're setting the defaults on the JSON schema properly.

tjasko (approved):
The changes look good! While I still think this [comment](https://github.com/Kong/koko/pull/490#discussion_r1005989708) is something we should consider if it's actually possible, the work required isn't trivial & we should punt it off into another PR anyway... also that it speeds up getting this released anyway.

PR #491, opened 2022-10-24
feat: process schema defaults before peforming schema validation

GGabriele:
Right now Koko requires a schema to include all required fields in order to pass validation via the `/v1/schemas/json/<entity>/validate` endpoint, regardless of these fields having schema defaults that are injected at run time.

For example:
```
$ echo '
id: 7f27869a-be49-406b-83f0-1c14becc1508
name: foo
host: example.com
path: /
' | y2j | http ":3000/v1/schemas/json/service/validate"
HTTP/1.1 400 Bad Request

{
    "code": 3,
    "details": [
        {
            "@type": "type.googleapis.com/kong.admin.model.v1.ErrorDetail",
            "messages": [
                "missing properties: 'protocol', 'port', 'connect_timeout', 'read_timeout', 'write_timeout'"
            ],
            "type": "ERROR_TYPE_ENTITY"
        }
    ],
    "message": "validation error"
}
```

This doesn't match the behaviour with Kong, which would let this payload pass because of the existing schema default values:

```
$ echo '
id: 7f27869a-be49-406b-83f0-1c14becc1508
name: foo
host: example.com
path: /
' | y2j | http ":8001/schemas/services/validate"
HTTP/1.1 200 OK

{
    "message": "schema validation successful"
}
```

Koko supports schemas defaults processing via the `ProcessDefaults` method. This commit makes sure defaults are injected before schema validation is performed.

### DIFF
--- a/internal/server/admin/schemas_service.go
+++ b/internal/server/admin/schemas_service.go
@@ -174,7 +174,10 @@ func (s *SchemasService) validateSchema(ctx context.Context, item proto.Message)
 		// registered (via `model.RegisterType()`), this would never error.
 		return err
 	}
-
+	// Process entity defaults before performing schema validation.
+	if err := obj.ProcessDefaults(ctx); err != nil {
+		return s.err(ctx, err)
+	}
 	return s.err(ctx, obj.Validate(ctx))
 }
 
--- a/internal/server/admin/schemas_test.go
+++ b/internal/server/admin/schemas_test.go
@@ -277,23 +277,17 @@ func TestValidateJSONSchema(t *testing.T) {
 
 		body := res.JSON().Object()
 		body.Value("message").String().Equal("validation error")
-		body.Value("details").Array().Length().Equal(4)
+		body.Value("details").Array().Length().Equal(3)
 		errRes := body.Value("details").Array()
 
-		entityErr := errRes.Element(0).Object()
-		entityErr.Value("type").String().Equal(v1.ErrorType_ERROR_TYPE_ENTITY.String())
-		messages := entityErr.Value("messages").Array()
-		messages.Length().Equal(1)
-		messages.First().String().Equal("missing properties: 'id'")
-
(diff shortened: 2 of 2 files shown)


aboudreault (approved):

internal/server/admin/schemas_service.go
```diff
@@ -174,7 +174,10 @@ func (s *SchemasService) validateSchema(ctx context.Context, item proto.Message)
 		// registered (via `model.RegisterType()`), this would never error.
 		return err
 	}
-
+	// Process entity defaults before performing schema validation.
+	if err := obj.ProcessDefaults(ctx); err != nil {
+		return err
```
tjasko, line 179: I have some work tracked to remove this silly error handling, but for now, we need to be very explicit:
```suggestion
		return s.err(ctx, err)
```

tjasko (changes requested):

internal/server/admin/schemas_service.go
GGabriele: done

aboudreault: for my understanding, what is the impact of not using that handler? 

I also see it is now used here as well: https://github.com/Kong/koko/pull/491/files#diff-f4d96c10298118aa8c7beeb59e9d437aa9438651e7bde8aebd6870d546ea19a1R175 (not part of this PR)

tjasko: If that handler isn't used in the applicable code path, the actual error will not be relayed to the end-user. E.g.: If that particular code exposes an error that end-users need to care about, it'd need to be wrapped with that error handler.

The code you linked @aboudreault wouldn't be an error an end-user needs to see.

tjasko (approved):

internal/server/admin/schemas_service.go
aboudreault: got it. thanks @tjasko!

PR #492, opened 2022-10-24
fix: no route defaults are applied if 'protocols' is used

GGabriele:
Right now, Koko doesn't apply routes' entity schema defaults when 'protocols' is not empty. The reason for this was probably an attempt to make schema validation easier: the `strip_path` default value is `true`, which is not compatible with `grpc` and `grpcs` protocols; when `protocols` is not empty defaults are not applied to prevent a incompatibility like the one about `strip_path` to occur.
This also prevents other defaults to be applied, which is not desiderable.

This commit removes this restriction by adding a check on the content of the `protocols` field, setting `strip_path` to `false` in the case it's not provided as an input.

### DIFF
--- a/internal/resource/route.go
+++ b/internal/resource/route.go
@@ -13,6 +13,7 @@ import (
 	"github.com/kong/koko/internal/model/json/generator"
 	"github.com/kong/koko/internal/model/json/validation"
 	"github.com/kong/koko/internal/model/json/validation/typedefs"
+	"github.com/samber/lo"
 	"google.golang.org/protobuf/types/known/wrapperspb"
 )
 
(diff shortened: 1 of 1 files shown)


internal/resource/route.go
```diff
@@ -126,13 +127,19 @@ func (r Route) ProcessDefaults(ctx context.Context) error {
...
 	}
-	if len(r.Route.Protocols) == 0 {
-		err := mergo.Merge(r.Route, defaultRoute,
-			mergo.WithTransformers(wrappersPBTransformer{}))
-		if err != nil {
-			return err
+	// When protocols has 'grpc' or 'grpcs', 'strip_path' cannot be set.
+	// If 'strip_path' is null, the default value (true) would be applied,
+	// so we set it to 'false' in order to prevent that.
+	if lo.Contains(r.Route.Protocols, "grpc") || lo.Contains(r.Route.Protocols, "grpcs") {
```
tjasko, line 133: There's a better way to do that w/ `lo` 😀
```suggestion
	if lo.Some(r.Route.Protocols, []string{typedefs.ProtocolGRPC, typedefs.ProtocolGRPCS}) {
```

tjasko, line 133: No test for this?

GGabriele: oh, nice!

GGabriele: It's already there: https://github.com/Kong/koko/blob/main/internal/resource/route_test.go#L901-L927

tjasko: Ah part of the validation test, that works w/ me.

tjasko (approved):

hbagdi:
Ouch. This conditional defaulting is very painful and causes a lot of trouble. We need to get rid of it.
No action is required in this PR.

PR #494, opened 2022-10-25
feat: add SslMode option to PostgresTLS config

aboudreault:
Add the ability to specify the tls connection ssl mode for postgresql. 

I do not validate the user value in koko because this is already done when parsing the dsn. e.g:
```
Error: database: unable to resolve DB migration driver from the given dialect: cannot parse `host=localhost port=5432 user=koko password=xxxxx dbname=koko sslmode=invalid sslrootcert=cluster.key`: failed to configure TLS (sslmode is invalid)
```

### DIFF
--- a/internal/config/db.go
+++ b/internal/config/db.go
@@ -59,6 +59,9 @@ type Postgres struct {
 type PostgresTLS struct {
 	CABundlePath string `yaml:"ca_bundle_path" json:"ca_bundle_path" env:"CA_BUNDLE_PATH"`
 	Enable       bool   `yaml:"enable" json:"enable" env:"ENABLE"`
+	// SslMode should be one of the postgresql ssl modes. Default to 'verify-full'.
+	// Ref: https://www.postgresql.org/docs/current/libpq-ssl.html#LIBPQ-SSL-SSLMODE-STATEMENTS
+	SslMode string `yaml:"ssl_mode" json:"ssl_mode" env:"SSL_MODE"`
 }
 
 // PostgresReadReplica allows for using a read replica in addition to the
@@ -109,6 +112,7 @@ func (c *Postgres) Opts() postgres.Opts {
 		CABundleFSPath:   c.TLS.CABundlePath,
 		DBName:           c.DBName,
 		EnableTLS:        c.TLS.Enable,
+		SslMode:          c.TLS.SslMode,
 		Port:             c.Port,
 		Hostname:         c.Hostname,
 		ReadOnlyHostname: c.ReadReplica.Hostname,
--- a/internal/persistence/postgres/postgres.go
+++ b/internal/persistence/postgres/postgres.go
@@ -19,7 +19,8 @@ const (
 	insertQuery = `insert into store(key,value) values($1,$2) on conflict (key) do update set value=$2;`
 	deleteQuery = `delete from store where key=$1`
 
-	DefaultPort = 5432
+	DefaultPort       = 5432
+	defaultTLSSslMode = "verify-full"
 )
 
 type Postgres struct {
@@ -37,6 +38,7 @@ type Opts struct {
 	Password         string
 	EnableTLS        bool
 	CABundleFSPath   string
+	SslMode          string
 	SQLOpen          persistence.SQLOpenFunc
 }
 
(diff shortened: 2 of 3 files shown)


internal/persistence/postgres/postgres.go
```diff
@@ -37,6 +38,7 @@ type Opts struct {
 	Password         string
 	EnableTLS        bool
 	CABundleFSPath   string
+	SslMode          string
```
mikefero, line 41: Should this be an enumeration or will an error occur is the `sslmode` is invalid for Postgres?

internal/persistence/postgres/postgres_test.go
```diff
@@ -40,6 +40,24 @@ func TestDSN(t *testing.T) {
...
+			CABundleFSPath: "/koko-postgres-cabundle/global-bundle.pem",
+			SslMode:        "verify-ca",
+		}
+		logger := log.Logger
+		expectedDSN := "host=localhost port=5432 user=koko password=koko dbname=koko sslmode=verify-ca " +
+			"sslrootcert=/koko-postgres-cabundle/global-bundle.pem"
+		dsn, err := getDSN(opt, logger)
+		require.NoError(t, err)
+		require.Equal(t, expectedDSN, dsn)
+	})
```
mikefero, line 60: We should add a test for invalid `SslMode` if an error is returned.

internal/persistence/postgres/postgres.go
aboudreault: yes, an error is returned by postgres: https://github.com/Kong/koko/pull/494#issue-1422502882

that's why I have not added an enum and did any koko validation. That option is only passed to the postgresql driver, so there is not much value doing so unless we really want to catch the config error at **config loading** instead of on dsn parsing+connection (do we?)

internal/persistence/postgres/postgres_test.go
aboudreault: The error will happen at connection, so I will check what we have for postgres integration tests.

internal/config/db.go
```diff
@@ -59,6 +59,9 @@ type Postgres struct {
 type PostgresTLS struct {
 	CABundlePath string `yaml:"ca_bundle_path" json:"ca_bundle_path" env:"CA_BUNDLE_PATH"`
 	Enable       bool   `yaml:"enable" json:"enable" env:"ENABLE"`
+	// SslMode should be one of the postgresql ssl modes. Default to 'verify-full'.
+	// Ref: https://www.postgresql.org/docs/current/libpq-ssl.html#LIBPQ-SSL-SSLMODE-STATEMENTS
+	SslMode string `yaml:"ssl_mode" json:"ssl_mode" env:"SSL_MODE"`
```
hbagdi, line 64: Why not give the user more control over all options using an "Options" string?

aboudreault: Interesting idea, you are thinking about **all** options, or just all TLS/SSL options? https://www.postgresql.org/docs/current/libpq-connect.html#LIBPQ-PARAMKEYWORDS

if we consider all options (so in `Postgres`, not `PostgresTLS`), those options would be appended to the DSN and passed to postgresql. Should we validate that the user doesn't specify any option that koko already generate, or we are good to just delegate this validation to postgres.

tjasko: A similar thing actually introduced with the MySQL changes, just didn't give the capability to Postgres. Though some care needed to be done, as certain options could break things:
https://github.com/Kong/koko/blob/8fdb6c3974453983299c92198e6c55c6bf5b06e7/internal/persistence/mysql/options.go#L20

tjasko, line 64: If we'd want to keep this, these fields should be named `SSLMode`:
https://github.com/golang/go/wiki/CodeReviewComments#initialisms

tjasko, line 64: You can use the `env-default:"verify-full"` struct tag to set the default btw.

hbagdi: Give users as many options as possible without making it easy for them to shoot themselves.

aboudreault: Let's go that direction then. I will come up with something similar to MySQL.

aboudreault:
closing in favor of https://github.com/Kong/koko/pull/500

PR #496, opened 2022-10-28
feat: zipkin traceid response header

omegabytes:
Adds `http_response_header_for_traceid`for zipkin.

See https://github.com/Kong/kong/pull/9173

### DIFF
--- a/internal/server/kong/ws/config/compat/compatibility_table.go
+++ b/internal/server/kong/ws/config/compat/compatibility_table.go
@@ -83,6 +83,7 @@ var (
 		"connect_timeout",
 		"send_timeout",
 		"read_timeout",
+		"http_response_header_for_traceid",
 	}
 	prometheus30Fields = []string{
 		"status_code_metrics",
--- a/internal/server/kong/ws/config/compat/compatibility_table.go
+++ b/internal/server/kong/ws/config/compat/compatibility_table.go
@@ -67,8 +67,8 @@ const (
 	versionsPre270      = "< 2.7.0"
 	versionsPre280      = "< 2.8.0"
 	versionsPre300      = "< 3.0.0"
+	versionsPre310      = "< 3.1.0"
 	versions300AndAbove = ">= 3.0.0"
-	versions310AndAbove = ">= 3.1.0"
 )
 
 var (
@@ -857,30 +857,16 @@ var (
 			Metadata: config.ChangeMetadata{
 				ID:       config.ChangeID("P136"),
 				Severity: config.ChangeSeverityError,
-				Description: "For the 'zipkin' plugin, " +
-					"'config.http_response_header_for_traceid' field is used " +
-					"in Kong Gateway versions >= 3.1. " +
-					"The plugin configuration has been updated to use " +
-					"'config.http_response_header_for_traceid' in the data-plane.",
-				Resolution: "Please update the configuration to use " +
-					"'config.http_response_header_for_traceid'.",
+				Description: standardPluginFieldsMessage("zipkin",
+					[]string{"http_response_header_for_traceid"},
+					"3.1", false),
+				Resolution: standardUpgradeMessage("3.1"),
 			},
-			SemverRange: versions310AndAbove,
+			SemverRange: versionsPre310,
 			Update: config.ConfigTableUpdates{
-				Name: "zipkin",
-				Type: config.Plugin,
-				FieldUpdates: []config.ConfigTableFieldCondition{
-					{
-						Field:     "http_response_header_for_traceid",
-						Condition: "http_response_header_for_traceid",
-						Updates: []config.ConfigTableFieldUpdate{
-							{
-								Field: "http_response_header_for_traceid",
-								Value: "X-B3-TraceId",
-							},
-						},
-					},
-				},
+				Name:         "zipkin",
+				Type:         config.Plugin,
+				RemoveFields: []string{"http_response_header_for_traceid"},
 			},
 		},
 	}
--- a/internal/server/kong/ws/config/compat/compatibility_table.go
+++ b/internal/server/kong/ws/config/compat/compatibility_table.go
@@ -68,6 +68,7 @@ const (
 	versionsPre280      = "< 2.8.0"
 	versionsPre300      = "< 3.0.0"
 	versions300AndAbove = ">= 3.0.0"
+	versionsPre310      = "< 3.1.0"
 )
 
 var (
@@ -83,7 +84,6 @@ var (
 		"connect_timeout",
 		"send_timeout",
 		"read_timeout",
-		"http_response_header_for_traceid",
 	}
 	prometheus30Fields = []string{
 		"status_code_metrics",
@@ -851,6 +851,22 @@ var (
 				Name:   config.Vault.String(),
 				Type:   config.Vault,
 				Remove: true,
+			},
+		},
+		{
+			Metadata: config.ChangeMetadata{
+				ID:       config.ChangeID("P136"),
+				Severity: config.ChangeSeverityError,
+				Description: standardPluginFieldsMessage("zipkin",
+					[]string{"http_response_header_for_traceid"},
+					"3.1", false),
+				Resolution: standardUpgradeMessage("3.1"),
+			},
+			SemverRange: versionsPre310,
+			Update: config.ConfigTableUpdates{
+				Name:         "zipkin",
+				Type:         config.Plugin,
+				RemoveFields: []string{"http_response_header_for_traceid"},
 			},
 		},
 	}
--- a/internal/server/kong/ws/config/compat/compatibility_table.go
+++ b/internal/server/kong/ws/config/compat/compatibility_table.go
@@ -68,7 +68,7 @@ const (
 	versionsPre280      = "< 2.8.0"
 	versionsPre300      = "< 3.0.0"
 	versions300AndAbove = ">= 3.0.0"
-	versionsPre310      = "< 3.1.0"
+	versions310AndAbove = ">= 3.1.0"
 )
 
 var (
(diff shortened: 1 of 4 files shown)


internal/server/kong/ws/config/compat/compatibility_table.go
```diff
@@ -75,6 +75,7 @@ var (
 		"connect_timeout",
 		"send_timeout",
 		"read_timeout",
+		"http_response_header_for_traceid",
```
mikefero, line 78: This is for 3.1 not 3.0; it will need to be defined in its own 3.1 version compatibility instance.

internal/test/e2e/plugins.go
```diff
@@ -405,7 +405,8 @@ var VersionCompatibilityOSSPluginConfigurationTests = []VersionCompatibilityPlug
 	//
 	// DP < 3.0:
 	//   - remove 'http_span_name', 'connect_timeout'
-	//     'send_timeout', 'read_timeout'
+	//     'send_timeout', 'read_timeout',
+	//     'http_response_header_for_traceid'
```
mikefero, line 409: Same call out here for the test. We should also create a test to ensure that this field (and eventually others) are going down to a 3.1 data plane which is currently [kong/kong:master-alpine](https://github.com/Kong/koko/blob/6846bca4d60159dc36b72fd645c0c86b028b5026/.github/workflows/integration.yaml#L16) executed during the integration GitHub Action.

omegabytes: > We should also create a test to ensure that this field (and eventually others) are going down to a 3.1 data plane

What does this mean? I provide an image tag when testing locally so I am hitting 3.1. Are you saying I also need to make a change to the CICD configuration?

omegabytes:
Integration Tests [OSS] / integration (kong/kong:master-alpine, false) (pull_request) failing is a known issue

internal/test/e2e/plugins.go
aboudreault: `kong/kong:master-alpine` is versioned `3.1.0`, so the CI is already testing the current status of the 3.1 release. We could probably add a test similar to where we use `versions300AndAbove` and make sure new fields are not altered by the compatibility stuff (so going down to a 3.1+ DP). 

@mikefero feel free to add more details in case I misunderstood this.

mikefero: Yep, that is exactly what I was referring to. Since this is for a new version of the Kong Gateway we want to ensure that this new field is being sent in the payload to these newer data planes. Validating these new releases is critical.

omegabytes: I've changed the logic to require the field for versions 3.1^ instead of removing the field for versions pre-3.1 [7f84e48](https://github.com/Kong/koko/pull/496/commits/7f84e48d7b13bf09330a71b673c8f1e1968893db).
In the [absence of documentation](https://docs.konghq.com/hub/kong-inc/zipkin/) on the field, I'm assuming we always want to set the value to `X-B3-TraceId` [based on the preceding work.](https://github.com/Kong/kong/pull/9173/files#diff-86d2d554cc8b36c294270788050e859dc26969f5d0495d888e56a27b764d1adfR478).

The reason I went with "remove for versions below 3.1" instead of "ensure presence for 3.1 and above" is because [this field is not required](https://github.com/Kong/kong/pull/9173/files#diff-4b0c5a7443b055f6888c3911c2ca1868b8fbebda607a49b52c9f266ec8f98859R68). Also, this probably should be a warning instead of an error, no?

internal/test/e2e/version_compatibility_test.go
```diff
@@ -211,7 +211,7 @@ func TestVersionCompatibility_PluginFieldUpdates(t *testing.T) {
 	for _, plugin := range expectedPluginsMap {
 		var config structpb.Struct
 		if len(plugin.Config) > 0 {
-			require.NoError(t, json.ProtoJSONUnmarshal([]byte(plugin.Config), &config))
+			require.NoError(t, json.ProtoJSONUnmarshal([]byte(plugin.Config), &config), plugin)
```
omegabytes, line 214: Adding this to the test fail message output helps isolate where an issue is. For example, I had a trailing comma, so the fail output was:
```
Error:      	Received unexpected error:
        	            	invalid character '}' looking for beginning of object key string
        	Test:       	TestVersionCompatibility_PluginFieldUpdates
        	Messages:   	{Name:zipkin Config:{
        	            				"local_service_name": "LOCAL_SERVICE_NAME",
        	            				"header_type": "ignore",
        	            				"http_span_name": "method_path",
        	            				"connect_timeout": 2001,
        	            				"send_timeout": 2001,
        	            				"read_timeout": 2001,
        	            			} Protocols:[] VersionRange: FieldUpdateChecks:map[< 2.7.0:[{Field:header_type Value:preserve}] >= 3.1.0:[{Field:http_response_header_for_traceid Value:X-B3-TraceId}]] ConfigureForService:true ConfigureForRoute:true ExpectedConfig:}
```

internal/test/e2e/plugins.go
aboudreault: @omegabytes , sorry, but I think there are some misunderstandings with the latest changes. Let me try to clarify:

1. The value of `http_response_header_for_traceid` can be set to anything the user want. This [line](https://github.com/Kong/kong/pull/9173/files#diff-86d2d554cc8b36c294270788050e859dc26969f5d0495d888e56a27b764d1adfR478) you pointed is only based on a test value, which is a simple [plugin configuration](https://github.com/Kong/kong/pull/9173/files#diff-86d2d554cc8b36c294270788050e859dc26969f5d0495d888e56a27b764d1adfR453). 

2. There is no `absence of documentation`, it is just not done yet. The gateway documentation is often done only 1-2 weeks before GA, depending on the workload.

3. We need a ChangeMetadata + test to ensure that if a user configure the zipkin plugin with that field, it is removed  for older DPs. At the moment, this is probably crashing (errors at config push in DP logs). This was OK in you previous version IIRC.

4. We want a test to ensure that if a user configure the zipkin plugin with that field, the plugin configuration is not altered by the version compatibility translator. It should not with your previous version, so it is just a sanity check that new fields of the schema for a version are correctly going down for the new 3.1 release. So that is exactly [this](https://github.com/Kong/koko/pull/496/files#diff-de20d2d66636a4ebbfeaf31f3cd4f7ef469b1fc14c7d2ab47a0e69efb4ba5191R428-R433) (if the plugin in test was configured with that field value)

internal/test/e2e/version_compatibility_test.go
GGabriele: Nice one!

internal/server/kong/ws/config/compat/compatibility_table.go
```diff
@@ -830,6 +831,36 @@ var (
...
+		{
+			Metadata: config.ChangeMetadata{
+				ID:       config.ChangeID("P135"),
+				Severity: config.ChangeSeverityError,
+				Description: "For the 'zipkin' plugin, " +
+					"'config.http_response_header_for_traceid' field is used " +
+					"in Kong Gateway versions >= 3.1. " +
+					"The plugin configuration has been updated to use " +
+					"'config.http_response_header_for_traceid' in the data-plane.",
+				Resolution: "Please update the configuration to use " +
```
GGabriele, line 843: I'm confused. Shouldn't we make sure that a new field is removed from payloads destined to  older DPs? This rule is making sure that `http_response_header_for_traceid` is set to `X-B3-TraceId` for newer DPs, but I think this should actually make sure to remove it for pre-3.1 DPs, basically a copy&paste of [this rule](https://github.com/Kong/koko/blob/main/internal/server/kong/ws/config/compat/compatibility_table.go#L700-L725) for the specific new config field.

omegabytes: image
I originally had a removal check (7f84e48) but I confused myself and removed it. I might add it back in after I go through [Alan's comment](https://github.com/Kong/koko/pull/496#discussion_r1010860736).

internal/test/e2e/plugins.go
```diff
@@ -423,6 +426,12 @@ var VersionCompatibilityOSSPluginConfigurationTests = []VersionCompatibilityPlug
 					Value: "preserve",
 				},
 			},
+			">= 3.1.0": {
```
GGabriele, line 429: Technically, this is not an update: the version compatibility rule is to remove this `http_response_header_for_traceid` field fold older DPs.

Adding things like this here to verify newer fields are present for newer DPs is not something we have been doing (for example, note that this is not checking that `read_timeout` or `send_timeout` are present for `>=3.0.0`.

I think we should remove this, and maybe add a new test making sure new fields are present for 3.1.0.

Thoughts?

aboudreault: +1, good point. Let's test this only when it's a `3.1.0` DP.  That way, we will also avoid modifying it if anything change in 3.2.0+. @omegabytes Note that we also want a test to check if that field is removed for <3.1 DPs.

aboudreault: ah nvm, I see it here: https://github.com/Kong/koko/pull/496/files#diff-de20d2d66636a4ebbfeaf31f3cd4f7ef469b1fc14c7d2ab47a0e69efb4ba5191R420 :+1:

omegabytes: [1869fc3](https://github.com/Kong/koko/pull/496/commits/1869fc3c2172cc7311650829f2b11c6512886655) asserts a valid config for 310 or newer. Includes some new fields as of 300 as well.

omegabytes:
`kong/kong:master-alpine` image is several weeks out of date. This causes the validation failure for the field we're adding in this branch. I've validated my changes locally against a kong image I've confirmed is up-to-date and contains the field changes we're adding here.

GGabriele (approved):

tjasko (approved):

aboudreault (approved):

PR #500, opened 2022-10-31
feat: add postgres config params

aboudreault:


internal/persistence/postgres/options.go
```diff
@@ -0,0 +1,111 @@
...
+			return fmt.Errorf(msg, key)
+		}
+	}
+
+	return nil
+}
+
+func (opts *Opts) DSN(logger *zap.Logger) (string, error) {
+	var dsn string
+	sslmode := defaultSSLMode
```
tjasko, line 61: This should ideally be named `sslMode`:
https://github.com/golang/go/wiki/CodeReviewComments#variable-names

internal/persistence/postgres/options.go
```diff
@@ -0,0 +1,111 @@
+package postgres
```
tjasko, line 1: Thanks for the refactor w/ copying what was done with the recent MySQL work. I didn't want to change the Postgres implementation with those changes, but figured we'd get to it eventually.

internal/persistence/postgres/options.go
```diff
@@ -0,0 +1,111 @@
...
+
+	if !opts.EnableTLS {
+		logger.Info("using non-TLS Postgres connection")
+		sslmode = "disable"
+	} else {
+		logger.Info("using TLS Postgres connection")
+		logger.Info("ca_bundle_fs_path:" + opts.CABundleFSPath)
+		if opts.CABundleFSPath == "" {
+			return "", fmt.Errorf("postgres connection requires TLS but ca_bundle_fs_path is empty")
+		}
```
tjasko, line 87: nit: You don't have to change anything here as it's existing code, and well, it's not directly related to this PR anyway.

But requiring this when TLS is enabled is weird, as it's an optional field:
https://github.com/jackc/pgx/blob/987de3874e0fc18b8c468d27cb9de9b2cd67540c/pgconn/config.go#L635-L638

Figured we might want to patch it up while we're at it, so figured to bring it up.

internal/persistence/postgres/options.go
```diff
@@ -0,0 +1,111 @@
...
+	if opts.DBName != "" {
+		dsn += fmt.Sprintf("dbname=%s", opts.DBName)
+	}
+
+	if !opts.EnableTLS {
+		logger.Info("using non-TLS Postgres connection")
+		sslmode = "disable"
+	} else {
+		logger.Info("using TLS Postgres connection")
+		logger.Info("ca_bundle_fs_path:" + opts.CABundleFSPath)
```
tjasko, line 84: nit: You don't have to change anything here as it's existing code.

Want to add a space after that colon?

internal/persistence/postgres/options.go
aboudreault: oops. my bad. :+1:

internal/persistence/postgres/options.go
```diff
@@ -0,0 +1,111 @@
...
+	}
+	if opts.User != "" {
+		dsn += fmt.Sprintf("user=%s ", opts.User)
+	}
+	if opts.Password != "" {
+		dsn += fmt.Sprintf("password=%s ", opts.Password)
+	}
+	if opts.DBName != "" {
+		dsn += fmt.Sprintf("dbname=%s", opts.DBName)
+	}
```
tjasko, line 77: nit: You don't have to change anything here as it's existing code.

This can actually be refactored to be a bit better, even when `pgx` doesn't expose any utils to form this config with typed fields. We could leverage [mitchellh/mapstructure](https://github.com/mitchellh/mapstructure) to label the options struct with the proper naming that `pgx` expects, convert it to a map, iterate upon those keys, and then spit out the proper config string using the standard `%v` formatted.

internal/persistence/postgres/options.go
aboudreault: good point. I'm for removing this now.

internal/persistence/postgres/options.go
aboudreault: +1

internal/persistence/postgres/options.go
aboudreault: I like the suggestion, but I would prefer to keep this as is since this change is require for another change planned this week. Also FYI, the reason I kept this code instead of adding them in the `params` map and passing everything to `formatDSNParams` is because I wanted to preserve the order of the base connection parameters can revisit this later for sure.

internal/persistence/postgres/options.go
aboudreault: done

internal/persistence/postgres/options.go
aboudreault: done

internal/persistence/postgres/options.go
aboudreault: done

internal/persistence/postgres/options.go
tjasko: Yup, absolutely. Definitely possible to do all this & preserve order, but I also concur it's a bit out-of-scope here.

tjasko (approved):
Looks good!

The commit description for https://github.com/Kong/koko/pull/500/commits/49673cb72051e501e6135cc8650079370f208693 is pointing to master, which that can change as time passes. Let's update the description to point to a specific commit ref or tag before merging.

aboudreault:
@tjasko good point, fixed! Could you reapprove the PR due to my force-push?

tjasko (approved):

tjasko (approved):

internal/persistence/postgres/options.go
```diff
@@ -0,0 +1,108 @@
...
+var (
+	defaultSSLMode            = "verify-full"
+	unsupportedParamErrPrefix = "the '%s' parameter is unsupported, "
+	unsupportedParamErrs      = map[string]string{
+		"host":        unsupportedParamErrPrefix + " please instead use 'postgres.hostname'",
+		"port":        unsupportedParamErrPrefix + " please instead use 'postgres.port'",
+		"dbname":      unsupportedParamErrPrefix + " please instead use 'postgres.db_name'",
+		"user":        unsupportedParamErrPrefix + " please instead use 'postgres.user'",
+		"password":    unsupportedParamErrPrefix + " please instead use 'postgres.password'",
+		"sslrootcert": unsupportedParamErrPrefix + " please instead use 'postgres.ca_bundle_path'",
```
mikefero, line 21: nit: remove `instead`

```suggestion
		"host":        unsupportedParamErrPrefix + " please use 'postgres.hostname'",
		"port":        unsupportedParamErrPrefix + " please use 'postgres.port'",
		"dbname":      unsupportedParamErrPrefix + " please use 'postgres.db_name'",
		"user":        unsupportedParamErrPrefix + " please use 'postgres.user'",
		"password":    unsupportedParamErrPrefix + " please use 'postgres.password'",
		"sslrootcert": unsupportedParamErrPrefix + " please use 'postgres.ca_bundle_path'",
```

internal/persistence/postgres/options.go
```diff
@@ -0,0 +1,108 @@
...
+type Opts struct {
+	DBName           string
+	Hostname         string
+	ReadOnlyHostname string
+	Port             int
+	User             string
+	Password         string
+	EnableTLS        bool
+	CABundleFSPath   string
+	SQLOpen          persistence.SQLOpenFunc
```
mikefero, line 34: Please make sure to add docs for this external options structure.

internal/persistence/postgres/options.go
aboudreault: done

internal/persistence/postgres/options.go
aboudreault: done

mikefero (approved):

PR #502, opened 2022-11-04
refactor: use pgxpool.Pool instead of database/sql for postgresql

aboudreault:
This PR is a refactor of the postgres persistence layer. Changes:

* Use the `pgxpool.Pool` interface to execute query instead of the built-in `database/sql`
* Add basic pooling options to the postgres config options.
* Make the postgres Pool plugable under the hood.

### DIFF
--- a/internal/config/config_test.go
+++ b/internal/config/config_test.go
@@ -2,8 +2,11 @@ package config
 
 import (
 	"testing"
+	"time"
 
 	"github.com/kong/koko/internal/db"
+	"github.com/kong/koko/internal/persistence"
+	"github.com/kong/koko/internal/persistence/postgres"
 	"github.com/stretchr/testify/require"
 )
 
@@ -19,6 +22,9 @@ var defaultConfig = Config{
 	Database: Database{
 		Dialect:      db.DialectSQLite3,
 		QueryTimeout: "5s",
+		Postgres: Postgres{
+			Pool: defaultPostgresPool,
+		},
 	},
 	Metrics: Metrics{
 		ClientType: "noop",
@@ -30,11 +36,23 @@ var defaultConfig = Config{
 	DisableAnonymousReports: false,
 }
 
+var defaultPostgresPool = PostgresPool{
+	Name:              postgres.DefaultPool,
+	MaxConns:          persistence.DefaultMaxConn,
+	MinConns:          persistence.DefaultMinConn,
+	MaxConnLifetime:   persistence.DefaultMaxConnLifetime,
+	MaxConnIdleTime:   persistence.DefaultMaxConnIdleTime,
+	HealthCheckPeriod: persistence.DefaultHealthCheckPeriod,
+}
+
 func TestGet(t *testing.T) {
 	type args struct {
 		filename string
 		envVars  map[string]string
 	}
+	duration20m, _ := time.ParseDuration("20m")
+	duration10m, _ := time.ParseDuration("10m")
+	duration30s, _ := time.ParseDuration("30s")
 	tests := []struct {
 		name      string
 		args      args
@@ -88,6 +106,14 @@ func TestGet(t *testing.T) {
 							Enable:       true,
 							CABundlePath: "/tmp/foo.crt",
 						},
+						Pool: PostgresPool{
+							Name:              "pgx",
+							MaxConns:          30,
+							MinConns:          5,
+							MaxConnLifetime:   duration20m,
+							MaxConnIdleTime:   duration10m,
+							HealthCheckPeriod: duration30s,
+						},
 					},
 					QueryTimeout: "2s",
 				},
@@ -137,6 +163,7 @@ func TestGet(t *testing.T) {
 							Enable:       true,
 							CABundlePath: "/tmp/foo.crt",
 						},
+						Pool: defaultPostgresPool,
 					},
 					QueryTimeout: "2s",
 				},
@@ -154,12 +181,13 @@ func TestGet(t *testing.T) {
 			name: "configuration can be provided via env vars",
 			args: args{
 				envVars: map[string]string{
-					"KOKO_LOG_LEVEL":                               "error",
-					"KOKO_LOG_FORMAT":                              "FOOBAR",
-					"KOKO_DATABASE_DIALECT":                        db.DialectPostgres,
-					"KOKO_DATABASE_POSTGRES_READ_REPLICA_HOSTNAME": "foobar",
-					"KOKO_DATABASE_POSTGRES_TLS_ENABLE":            "true",
-					"KOKO_METRICS_PROMETHEUS_ENABLE":               "true",
+					"KOKO_LOG_LEVEL":                                "error",
+					"KOKO_LOG_FORMAT":                               "FOOBAR",
+					"KOKO_DATABASE_DIALECT":                         db.DialectPostgres,
+					"KOKO_DATABASE_POSTGRES_READ_REPLICA_HOSTNAME":  "foobar",
+					"KOKO_DATABASE_POSTGRES_TLS_ENABLE":             "true",
+					"KOKO_DATABASE_POSTGRES_POOL_MAX_CONN_LIFETIME": "20m",
+					"KOKO_METRICS_PROMETHEUS_ENABLE":                "true",
 				},
 			},
 			want: Config{
@@ -178,6 +206,14 @@ func TestGet(t *testing.T) {
 						},
 						TLS: PostgresTLS{
 							Enable: true,
+						},
+						Pool: PostgresPool{
+							Name:              postgres.DefaultPool,
+							MaxConns:          persistence.DefaultMaxConn,
+							MinConns:          persistence.DefaultMinConn,
+							MaxConnLifetime:   duration20m,
+							MaxConnIdleTime:   persistence.DefaultMaxConnIdleTime,
+							HealthCheckPeriod: persistence.DefaultHealthCheckPeriod,
 						},
 					},
 					QueryTimeout: "5s",
@@ -232,6 +268,7 @@ func TestGet(t *testing.T) {
 							Enable:       true,
 							CABundlePath: "/tmp/foo.crt",
 						},
+						Pool: defaultPostgresPool,
 					},
 					QueryTimeout: "2s",
 				},
--- a/internal/config/db.go
+++ b/internal/config/db.go
@@ -53,6 +53,7 @@ type Postgres struct {
 	TLS         PostgresTLS         `yaml:"tls" json:"tls" env-prefix:"TLS_"`
 	User        string              `yaml:"user" json:"user" env:"USER"`
 	Password    string              `yaml:"password" json:"password" env:"PASSWORD"`
+	Pool        PostgresPool        `yaml:"pool" json:"pool" env-prefix:"POOL_"`
 
 	// See the `Params` field on postgres.Opts for more info.
 	Params map[string]string `yaml:"params" json:"params" env:"PARAMS"`
@@ -68,6 +69,16 @@ type PostgresTLS struct {
 // primary which shares the same connection settings as the primary DB.
 type PostgresReadReplica struct {
 	Hostname string `yaml:"hostname" json:"hostname" env:"HOSTNAME"`
+}
+
+// PostgresPool defines configuration for connections pooling.
+type PostgresPool struct {
+	Name              string        `yaml:"name" json:"name" env:"NAME" env-default:"pgx"`
+	MaxConns          int           `yaml:"max_connections" json:"max_connections" env:"MAX_CONNS" env-default:"50"`
+	MinConns          int           `yaml:"min_connections" json:"min_connections" env:"MIN_CONNS" env-default:"2"`
+	MaxConnLifetime   time.Duration `yaml:"max_connection_lifetime" json:"max_connection_lifetime" env:"MAX_CONN_LIFETIME" env-default:"1h"`    //nolint:lll
+	MaxConnIdleTime   time.Duration `yaml:"max_connection_idle_time" json:"max_connection_idle_time" env:"MAX_CONN_IDLE_TIME" env-default:"1h"` //nolint:lll
+	HealthCheckPeriod time.Duration `yaml:"health_check_period" json:"health_check_period" env:"HEALTH_CHECK_PERIOD" env-default:"1m"`          //nolint:lll
 }
 
 // SQLite defines configuration for using SQLite as the persistent store.
@@ -103,7 +114,7 @@ func (c *MySQL) Opts() (mysql.Opts, error) {
 }
 
 // Opts returns the options required to instantiate a Postgres persistence.Persister.
-func (c *Postgres) Opts() postgres.Opts {
+func (c *Postgres) Opts() (postgres.Opts, error) {
 	if c.Port == 0 {
 		c.Port = postgres.DefaultPort
 	}
@@ -113,11 +124,23 @@ func (c *Postgres) Opts() postgres.Opts {
 		DBName:           c.DBName,
 		EnableTLS:        c.TLS.Enable,
 		Params:           c.Params,
+		Pool:             c.poolOpts(),
 		Port:             c.Port,
 		Hostname:         c.Hostname,
 		ReadOnlyHostname: c.ReadReplica.Hostname,
 		User:             c.User,
 		Password:         c.Password,
+	}, nil
+}
+
+func (c *Postgres) poolOpts() postgres.PoolOpts {
+	return postgres.PoolOpts{
+		Name:              c.Pool.Name,
+		MaxConns:          c.Pool.MaxConns,
+		MinConns:          c.Pool.MinConns,
+		MaxConnLifetime:   c.Pool.MaxConnLifetime,
+		MaxConnIdleTime:   c.Pool.MaxConnIdleTime,
+		HealthCheckPeriod: c.Pool.HealthCheckPeriod,
 	}
 }
 
@@ -146,12 +169,17 @@ func ToDBConfig(config Database, logger *zap.Logger) (db.Config, error) {
 		return db.Config{}, fmt.Errorf("unable to set MySQL DB options: %w", err)
 	}
 
+	postgresOpts, err := config.Postgres.Opts()
+	if err != nil {
+		return db.Config{}, fmt.Errorf("unable to set Postgres DB options: %w", err)
+	}
+
 	return db.Config{
 		Dialect:      config.Dialect,
 		QueryTimeout: queryTimeout,
 		Logger:       logger,
 		MySQL:        mysqlOpts,
-		Postgres:     config.Postgres.Opts(),
+		Postgres:     postgresOpts,
 		SQLite:       config.SQLite.Opts(),
 	}, nil
 }
--- a/internal/persistence/driver.go
+++ b/internal/persistence/driver.go
@@ -7,10 +7,13 @@ import (
 
 // Various default settings used for non-SQLite DB drivers.
 const (
-	DefaultDialTimeout     = 15 * time.Second
-	DefaultMaxConn         = 50
-	DefaultMaxConnLifetime = time.Hour
-	DefaultMaxIdleConn     = 20
+	DefaultDialTimeout       = 15 * time.Second
+	DefaultMaxConn           = 50
+	DefaultMinConn           = 2
+	DefaultMaxConnLifetime   = time.Hour
+	DefaultMaxConnIdleTime   = time.Hour
+	DefaultMaxIdleConn       = 20
+	DefaultHealthCheckPeriod = time.Minute
 )
 
 // SQLOpenFunc defines a function that can instantiate a sql.DB instance from the given DSN.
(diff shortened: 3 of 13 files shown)


internal/persistence/postgres/postgres.go
```diff
@@ -74,41 +83,43 @@ func (s *Postgres) Driver() persistence.Driver { return persistence.Postgres }
...
 	error,
 ) {
-	q := postgresQuery{query: s.readOnlyDB, queryTimeout: s.queryTimeout}
+	q := postgresQuery{query: s.readOnlyDBPool, queryTimeout: s.queryTimeout}
 	return q.List(ctx, prefix, opts)
 }
 
 func (s *Postgres) Tx(ctx context.Context) (persistence.Tx, error) {
-	tx, err := s.db.BeginTx(ctx, nil)
+	tx, err := s.dbPool.BeginTx(ctx, pgx.TxOptions{})
```
aboudreault, line 116: With `database/sql`, a transaction will be automatically roll backed on context cancellation. This is not something we have by default with `pgx`. Although we don't have any test that is testing this scenario, I think it would be a good to do the same thing here, wdyt?

internal/persistence/postgres/pool.go
```diff
@@ -0,0 +1,154 @@
...
+
+// PoolOpenFunc defines a function that can instantiate a Pool instance with the given options.
+type PoolOpenFunc func(opts Opts, logger *zap.Logger) (Pool, error)
+
+type PoolOpts struct {
+	// Optional The name of the postgres pool implementation to use.
+	// By default, only `pgx` is available.
+	Name string
+
+	// MaxConns is the maximum size of the pool. Default to `persistence.DefaultMaxConn`.
```
aboudreault, line 48: Currently, this is a bit falsy. The default are only set in the config yaml stuff...

internal/persistence/postgres/pool.go
```diff
@@ -0,0 +1,154 @@
...
+	// closed by the health check. Default to `persistence.DefaultMaxConnLifetime`.
+	MaxConnIdleTime time.Duration
+
+	// HealthCheckPeriod is the duration between checks of the health of
+	// idle connections. Default to `persistence.DefaultHealthCheckPeriod`.
+	HealthCheckPeriod time.Duration
+
+	// ReadOnly indicates the pool will be used for read only operations.
+	ReadOnly bool
+}
```
aboudreault, line 68: Building a new PoolOpts  using `PoolOpts{}` will give us an invalid default object, due to the zero-value of these types (see PoolOpts.Validate()). This is not ideal. Should I create a proper `newPooOpts()` to avoid this? (not sure yet how this would work since we also use `PostgresOpts{}` directly.

internal/persistence/postgres/pool.go
```diff
@@ -0,0 +1,154 @@
...
+	if _, present := pools[poolName]; present {
+		return fmt.Errorf("pool name '%s' already exists", poolName)
+	}
+	pools[poolName] = poolOpenFunc
+
+	return nil
+}
+
+func init() {
+	pools[DefaultPool] = newPgxPool
```
aboudreault, line 153: use `RegisterPool()` and handle the error with a simple panic.

aboudreault:
@Kong/koko-engineering this is ready for a first review, if anyone has some free cycles

internal/config/db.go
```diff
@@ -70,6 +71,16 @@ type PostgresReadReplica struct {
...
 
+// PostgresPool defines configuration for connections pooling.
+type PostgresPool struct {
+	Name              string `yaml:"name" json:"name" env:"NAME" env-default:"pgx"`
+	MaxConns          int32  `yaml:"max_conns" json:"max_conns" env:"MAX_CONNS" env-default:"50"`
+	MinConns          int32  `yaml:"min_conns" json:"min_conns" env:"MIN_CONNS" env-default:"2"`
+	MaxConnLifetime   string `yaml:"max_conn_lifetime" json:"max_conn_lifetime" env:"MAX_CONN_LIFETIME" env-default:"1h"`
+	MaxConnIdleTime   string `yaml:"max_conn_idle_time" json:"max_conn_idle_time" env:"MAX_CONN_IDLE_TIME" env-default:"1h"`    //nolint:lll
+	HealthCheckPeriod string `yaml:"health_check_period" json:"health_check_period" env:"HEALTH_CHECK_PERIOD" env-default:"1m"` //nolint:lll
+}
```
omegabytes, line 82: How did we choose these as default values?

aboudreault: these defaults come from koko if they were existing, otherwise pgx itself.

internal/persistence/postgres/pool.go
aboudreault: yeah, I wrote this comment prior the addition of the pool validation function: https://github.com/Kong/koko/pull/502/files#diff-d18f5e36820a1a2d869d4c377f9d3bc3ee07bd24391f4c43ec7e7a6b80e1b5f9R71

so with the default yaml settings being applied at config loading.. and also koko calling this Validate function, the normal path is fine.

internal/persistence/postgres/options.go
```diff
@@ -53,6 +53,9 @@ type Opts struct {
 	// information about connection parameters.
 	// See: https://www.postgresql.org/docs/current/runtime-config.html for more information about runtime parameters.
 	Params map[string]string
+
+	// Optional defines configuration for connections pooling.
```
GGabriele, line 57:
```suggestion
	// Pool defines configuration for connections pooling.
```

internal/persistence/postgres/pool.go
```diff
@@ -0,0 +1,154 @@
...
+
+	// MinConns is the minimum size of the pool. Default to `persistence.DefaultMinConn`.
+	MinConns int32
+
+	// MaxConnLifetime is the duration since creation after which a connection will
+	// be automatically closed. Default to `persistence.DefaultMaxConnLifetime`.
+	MaxConnLifetime time.Duration
+
+	// MaxConnIdleTime is the duration after which an idle connection will be automatically
+	// closed by the health check. Default to `persistence.DefaultMaxConnLifetime`.
```
GGabriele, line 59:
```suggestion
	// closed by the health check. Default to `persistence.MaxConnIdleTime`.
```

internal/persistence/postgres/pool_test.go
```diff
@@ -0,0 +1,120 @@
...
+				Pool:     poolOpt,
+			}
+
+			err := opt.Validate()
+			require.Error(t, err)
+		}
+	})
+}
+
+func TestPoolRegistration(t *testing.T) {
```
GGabriele, line 75: Maybe add a test case for an already registered pool?

aboudreault: +1, will do

internal/persistence/postgres/pool.go
```diff
@@ -0,0 +1,154 @@
...
+	config.MaxConnLifetime = opts.Pool.MaxConnLifetime
+	config.MaxConnIdleTime = opts.Pool.MaxConnIdleTime
+	config.HealthCheckPeriod = opts.Pool.HealthCheckPeriod
+
+	return config, nil
+}
+
+func newPgxPool(opts Opts, logger *zap.Logger) (Pool, error) {
+	logger.Info("using Postgres PGX Pool")
+	ctx := context.Background()
```
tjasko, line 122: Let's use a context with a hard-coded timeout of like 15s or so, because we wouldn't want this to hang in the event it can't establish the connection or ping in a decent timeframe.

internal/persistence/postgres/pool.go
```diff
@@ -0,0 +1,154 @@
...
+
+	dbPool, err := pgxpool.ConnectConfig(ctx, pgxPoolConfig)
+	if err != nil {
+		return nil, err
+	}
+
+	err = dbPool.Ping(ctx)
+	if err != nil {
+		return nil, err
+	}
```
tjasko, line 137: Why is this happening? `pgxpool.ConnectConfig()` will already establish the connection, so this ping seems unnecessary, no?

internal/persistence/postgres/pool.go
```diff
@@ -0,0 +1,154 @@
...
+	if _, present := pools[poolName]; present {
+		return fmt.Errorf("pool name '%s' already exists", poolName)
+	}
+	pools[poolName] = poolOpenFunc
+
+	return nil
+}
+
+func init() {
+	pools[DefaultPool] = newPgxPool
```
tjasko, line 153: I don't understand why we're doing this. Can we throw some documentation on why we need to store all the registered pools? I imagine it's for the healthcheck stuff to ensure we can easily look up all pools at once, but perhaps I'm wrong.

internal/persistence/postgres/pool.go
```diff
@@ -0,0 +1,154 @@
...
+	// MaxConnIdleTime is the duration after which an idle connection will be automatically
+	// closed by the health check. Default to `persistence.DefaultMaxConnIdleTime`.
+	MaxConnIdleTime time.Duration
+
+	// HealthCheckPeriod is the duration between checks of the health of
+	// idle connections. Default to `persistence.DefaultHealthCheckPeriod`.
+	HealthCheckPeriod time.Duration
+
+	// ReadOnly indicates the pool will be used for read only operations.
+	ReadOnly bool
```
tjasko, line 67: How will this be used, as it's currently unused?

internal/config/db.go
```diff
@@ -70,6 +71,16 @@ type PostgresReadReplica struct {
...
 }
 
+// PostgresPool defines configuration for connections pooling.
+type PostgresPool struct {
+	Name              string `yaml:"name" json:"name" env:"NAME" env-default:"pgx"`
+	MaxConns          int32  `yaml:"max_conns" json:"max_conns" env:"MAX_CONNS" env-default:"50"`
+	MinConns          int32  `yaml:"min_conns" json:"min_conns" env:"MIN_CONNS" env-default:"2"`
+	MaxConnLifetime   string `yaml:"max_conn_lifetime" json:"max_conn_lifetime" env:"MAX_CONN_LIFETIME" env-default:"1h"`
+	MaxConnIdleTime   string `yaml:"max_conn_idle_time" json:"max_conn_idle_time" env:"MAX_CONN_IDLE_TIME" env-default:"1h"`    //nolint:lll
+	HealthCheckPeriod string `yaml:"health_check_period" json:"health_check_period" env:"HEALTH_CHECK_PERIOD" env-default:"1m"` //nolint:lll
```
tjasko, line 81: This should probably be disabled by default, right?

internal/persistence/postgres/tx.go
```diff
@@ -0,0 +1,42 @@
+package postgres
+
+import (
+	"context"
+
+	"github.com/jackc/pgx/v4"
+	"github.com/kong/koko/internal/persistence"
+)
+
+type postgresTx struct {
```
tjasko, line 10: Thanks for pulling that out like I did with MySQL, def helps keep the code more readable.

tjasko:
Great work here @aboudreault 😄

internal/persistence/postgres/postgres.go
aboudreault: ACK: thanks!

internal/persistence/postgres/pool.go
```diff
@@ -0,0 +1,154 @@
...
+	"github.com/jackc/pgx/v4/pgxpool"
+	"go.uber.org/zap"
+)
+
+var pools = map[string]PoolOpenFunc{}
+
+// Pool is the interface of the pgxpool.Pool type, used by the postgres
+// persistence layer to query the database. A custom pool type can be
+// registered using `RegisterPool` if it implements this interface.
+type Pool interface {
```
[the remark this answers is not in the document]
aboudreault: since `*pgxpool.Pool` is a type, I cannot use it as the abstraction type/interface for my different pool implementations. e.g. My custom aurora pool implementation is returning a `*AuroraPool`. I need a way that our postgres persistence layer can rely on and just call the query execution functions. does that make sense or I am missing something?

internal/persistence/postgres/pool.go
aboudreault: currently, this is used in the aurora impl to determine what kind of healthcheck validator is required (read vs write). However, we could remove that if we had a generic `PoolOpts.params` that we can pass thru.

internal/persistence/postgres/pool.go
aboudreault: will fix!

internal/persistence/postgres/pool.go
```diff
@@ -0,0 +1,154 @@
...
+func newPostgresPool(opts Opts, logger *zap.Logger) (Pool, error) {
+	var dbPool Pool
+	var err error
+
+	if poolOpenFunc, present := pools[opts.Pool.Name]; present {
+		dbPool, err = poolOpenFunc(opts, logger)
+		if err != nil {
+			return nil, err
+		}
+		return dbPool, nil
```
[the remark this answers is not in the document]
aboudreault: indeed, +1!!

internal/persistence/postgres/pool.go
```diff
@@ -0,0 +1,154 @@
...
+	config.MinConns = opts.Pool.MinConns
+	config.MaxConnLifetime = opts.Pool.MaxConnLifetime
+	config.MaxConnIdleTime = opts.Pool.MaxConnIdleTime
+	config.HealthCheckPeriod = opts.Pool.HealthCheckPeriod
+
+	return config, nil
+}
+
+func newPgxPool(opts Opts, logger *zap.Logger) (Pool, error) {
+	logger.Info("using Postgres PGX Pool")
```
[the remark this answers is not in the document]
aboudreault: will fix!

internal/persistence/postgres/pool.go
aboudreault: good point, will fix!

internal/persistence/postgres/pool.go
aboudreault: umm, yeah let's remove this.

internal/persistence/postgres/pool.go
```diff
@@ -0,0 +1,154 @@
...
+	err = dbPool.Ping(ctx)
+	if err != nil {
+		return nil, err
+	}
+
+	return dbPool, nil
+}
+
+// RegisterPool adds a `Pool` implementation to the postgres configuration.
+func RegisterPool(poolName string, poolOpenFunc PoolOpenFunc) error {
```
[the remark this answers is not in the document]
aboudreault: +1, will fix

internal/persistence/postgres/pool.go
aboudreault: not what what you mean, it's only to be able to select the proper pool implementation by name from the configuration. e.g:
`  postgres:
    hostname: localhost
    port: 5432
    db_name: koko
    user: koko
    password: koko
    pool:
      name: aurora
      health_check_period: 10s
`

so a `pool_name:pool_init_func` registration.

internal/persistence/postgres/postgres.go
```diff
@@ -74,41 +83,43 @@ func (s *Postgres) Driver() persistence.Driver { return persistence.Postgres }
 
 // SetDefaultSQLOptions implements the persistence.SQLPersister interface.
 func (s *Postgres) SetDefaultSQLOptions(db *sql.DB) error {
+	// default settings are set here because the standard dbgit  client is only used for migrations.
```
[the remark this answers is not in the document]
aboudreault: totally!

aboudreault (commented):
Thanks Taylor for all your comments! I responded and will fix a bunch of them tomorrow morning.

internal/persistence/postgres/pool_test.go
aboudreault: done

internal/persistence/postgres/postgres.go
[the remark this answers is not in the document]
aboudreault: done

internal/config/config.go
```diff
@@ -29,3 +29,5 @@ func Get(filename string) (Config, error) {
 	}
 	return c, nil
 }
+
+// func Read()
```
rajkong, line 33: Do we need this?

aboudreault: not at all!! will remove

internal/config/config_test.go
```diff
@@ -30,6 +35,15 @@ var defaultConfig = Config{
 	DisableAnonymousReports: false,
 }
 
+var defaultPostgresPool = PostgresPool{
+	Name:              postgres.DefaultPool,
+	MaxConns:          persistence.DefaultMaxConn,
+	MinConns:          persistence.DefaultMinConn,
+	MaxConnLifetime:   "1h",
```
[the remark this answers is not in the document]
aboudreault: done

internal/config/config.go
aboudreault: done

internal/persistence/driver.go
```diff
@@ -7,10 +7,13 @@ import (
 
 // Various default settings used for non-SQLite DB drivers.
 const (
-	DefaultDialTimeout     = 15 * time.Second
-	DefaultMaxConn         = 50
-	DefaultMaxConnLifetime = time.Hour
-	DefaultMaxIdleConn     = 20
+	DefaultDialTimeout       = 15 * time.Second
```
rajkong, line 10: Will driver.go change when users want to add implementations?

internal/persistence/postgres/pool.go
rajkong: I am not sure about the value of the ReadOnly parameter as well. Perhaps this is something to talk about internally?

aboudreault: sure Rajan. Here is how the ro pool is currently created: https://github.com/Kong/koko/pull/502/files#diff-f39571bb34baebee68e4a03486bb984084067ec6ac049be19f648184888d2494R63-R71

feel free to ping me on on Slack and/or suggest an alternative way to do this. we would also add a readOnly argument to ` newPostgresPool` but not sure this is cleaner anyway.

internal/persistence/driver.go
aboudreault: that `SQLOpenFunc` is not used for pool initialization.. I should add something in its docs. maybe you can elaborate on your thoughts about it offline.

internal/persistence/postgres/pool.go
aboudreault: this was discussed offline with Rajan and we agreed it was indeed the simpler way to accomplish what we need in the custom pool.

rajkong: Discussed and the approach is ok

rajkong (commented):
Looks good to me. Waiting for other comments to be resolved before approval

internal/config/db.go
```diff
@@ -70,6 +71,16 @@ type PostgresReadReplica struct {
 	Hostname string `yaml:"hostname" json:"hostname" env:"HOSTNAME"`
 }
 
+// PostgresPool defines configuration for connections pooling.
+type PostgresPool struct {
+	Name              string `yaml:"name" json:"name" env:"NAME" env-default:"pgx"`
```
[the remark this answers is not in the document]
tjasko: Ah, got it, wasn't thinking there could be multiple pool implementations (thought that wouldn't been done differently).

internal/config/db.go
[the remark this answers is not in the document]
tjasko: Err, I linked to the wrong lib lol. Sorry, the library we're using for config unmarshalling also supports this:
https://github.com/ilyakaznacheev/cleanenv

We just need to change the type to `time.Duration`, and it'll just work as expected.

internal/config/db.go
```diff
@@ -70,6 +71,16 @@ type PostgresReadReplica struct {
...
+	MaxConns          int32  `yaml:"max_conns" json:"max_conns" env:"MAX_CONNS" env-default:"50"`
+	MinConns          int32  `yaml:"min_conns" json:"min_conns" env:"MIN_CONNS" env-default:"2"`
```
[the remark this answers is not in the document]
tjasko: Yeah that's perfectly fine to cast when appropriate (as after all, pgx is a third-party lib). Thanks!

internal/persistence/postgres/pool.go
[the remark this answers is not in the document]
tjasko: Cool, thanks!

internal/persistence/postgres/pool.go
tjasko: Err, yeah you can forget this, I see what it's doing now. I should have deleted this comment.

internal/config/db.go
tjasko: Ohh, completely sidestepped on the fact that the pgx pool implementation implements its own health check too... derp, mind me.

tjasko:
I dropped the ball on this @aboudreault, I apologize for the delay getting back to you. Other than the below comment, all looks good!
https://github.com/Kong/koko/pull/502#discussion_r1037610001

internal/config/db.go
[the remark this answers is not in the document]
aboudreault: ah! no idea why I thought Duration was supported in cleanenv without even looking :facepalm: ... maybe because of how `queryTimeout` was handled. fixed!

tjasko (approved):

PR #509, opened 2022-11-11
fix: Disable Change Tracking for Zipkin TraceID Header

omegabytes:
Adds a check on config.http_response_header_for_traceid for the zipkin plugin. Change tracking is conditionally ignored when this field is set to the default value (nil).

### DIFF
--- a/internal/server/kong/ws/config/compat/compatibility_table_test.go
+++ b/internal/server/kong/ws/config/compat/compatibility_table_test.go
@@ -925,6 +925,7 @@ func TestDisableChangeTracking(t *testing.T) {
 					"connect_timeout": 2000,
 					"read_timeout": 5000,
 					"send_timeout": 5000,
+					"http_response_header_for_traceid": null
 				}
 			}
 		]
@@ -1138,6 +1139,53 @@ func TestDisableChangeTracking(t *testing.T) {
 				ChangeDetails: []config.ChangeDetail{
 					{
 						ID: "P116",
+						Resources: []config.ResourceInfo{
+							{
+								Type: "plugin",
+								ID:   "759c0d3a-bc3d-4ccc-8d4d-f92de95c1f1a",
+							},
+						},
+					},
+				},
+			},
+		},
+		{
+			name: "[zipkin] change is emitted with non-default value for" +
+				" http_response_header_for_traceid",
+			uncompressedPayload: `
+{
+	"config_table": {
+		"plugins": [
+			{
+				"id": "759c0d3a-bc3d-4ccc-8d4d-f92de95c1f1a",
+				"name": "zipkin",
+				"config": {
+					"http_response_header_for_traceid": "X-B3-TraceId"
+				}
+			}
+		]
+	}
+}
+`,
+			dataPlaneVersion: "2.7.0",
+			expectedPayload: `
+{
+	"config_table": {
+		"plugins": [
+			{
+				"id": "759c0d3a-bc3d-4ccc-8d4d-f92de95c1f1a",
+				"name": "zipkin",
+				"config": {
+				}
+			}
+		]
+	}
+}
+`,
+			expectedChanges: config.TrackedChanges{
+				ChangeDetails: []config.ChangeDetail{
+					{
+						ID: "P136",
 						Resources: []config.ResourceInfo{
 							{
 								Type: "plugin",
--- a/internal/server/kong/ws/config/compat/compatibility_table.go
+++ b/internal/server/kong/ws/config/compat/compatibility_table.go
@@ -867,6 +867,11 @@ var (
 				Name:         "zipkin",
 				Type:         config.Plugin,
 				RemoveFields: []string{"http_response_header_for_traceid"},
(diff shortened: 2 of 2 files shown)


mikefero:
Please add a test for this to ensure that track changes is not emitted when the default is set.

internal/server/kong/ws/config/compat/compatibility_table_test.go
```diff
@@ -1148,6 +1149,53 @@ func TestDisableChangeTracking(t *testing.T) {
 				},
 			},
 		},
+		{
```
mikefero, line 1152: Please also add the test for ensuring that it is not emitted.

omegabytes: @mikefero that is this https://github.com/Kong/koko/pull/509/files#diff-4a012eaac7d341b5a8d54daa458e5091b0d3ceb450237361164a17392134ced6R928

internal/server/kong/ws/config/compat/compatibility_table_test.go
```diff
@@ -1148,6 +1149,53 @@ func TestDisableChangeTracking(t *testing.T) {
...
+			name: "[zipkin] change is emitted with non-default value for" +
+				" http_response_header_for_traceid",
```
mikefero, line 1154:
```suggestion
			name: "[zipkin] change is not emitted with default value for" +
				" http_response_header_for_traceid",
```

internal/server/kong/ws/config/compat/compatibility_table_test.go
mikefero: I forgot to hit submit on a change request (your name was wrong for this test); this test is not emitted, but we also need one to ensure it is emitted. This is separate from the integration tests and requires a unit test too.

internal/server/kong/ws/config/compat/compatibility_table_test.go
omegabytes: The default value for `http_response_header_for_traceid` is `null`. The test "`[zipkin] change is emitted with non-default value for http_response_header_for_traceid`" asserts that the upgrade warning is emitted when `http_response_header_for_traceid`, set to a non-default value, is dropped from the config for a incompatible data plane version.

omegabytes: If I update this test by removing the [expected change](https://github.com/Kong/koko/blob/71ff2bb014f148ca0cfded61041448701c8612a4/internal/server/kong/ws/config/compat/compatibility_table.go#L856-L876), we get an error:
```
        	Error:      	Not equal: 
        	            	expected: config.TrackedChanges{ChangeDetails:[]config.ChangeDetail(nil)}
        	            	actual  : config.TrackedChanges{ChangeDetails:[]config.ChangeDetail{config.ChangeDetail{ID:"P136", Resources:[]config.ResourceInfo{config.ResourceInfo{Type:"plugin", ID:"759c0d3a-bc3d-4ccc-8d4d-f92de95c1f1a"}}}}}
        	            	
        	            	Diff:
        	            	--- Expected
        	            	+++ Actual
        	            	@@ -1,3 +1,13 @@
        	            	 (config.TrackedChanges) {
        	            	- ChangeDetails: ([]config.ChangeDetail) <nil>
        	            	+ ChangeDetails: ([]config.ChangeDetail) (len=1) {
        	            	+  (config.ChangeDetail) {
        	            	+   ID: (config.ChangeID) (len=4) "P136",
        	            	+   Resources: ([]config.ResourceInfo) (len=1) {
        	            	+    (config.ResourceInfo) {
        	            	+     Type: (string) (len=6) "plugin",
        	            	+     ID: (string) (len=36) "759c0d3a-bc3d-4ccc-8d4d-f92de95c1f1a"
        	            	+    }
        	            	+   }
        	            	+  }
        	            	+ }
        	            	 }
Test:       	TestDisableChangeTracking/[zipkin]_change_is_emitted_with_non-default_value_for_http_response_header_for_traceid
```

mikefero: My bad, yes you are right there; please ignore the suggestion. You do need to add a test though to make sure it is emitted.

omegabytes: This PR already adds `http_response_header_for_traceid: null` to the [existing default value test](https://github.com/Kong/koko/blob/71ff2bb014f148ca0cfded61041448701c8612a4/internal/server/kong/ws/config/compat/compatibility_table_test.go#L914-L951). This asserts that a change message is not emitted when `http_response_header_for_traceid` set to the default value is dropped from a config for an incompatible data plane. 

Are you requesting that I add a test that asserts that a change message is not emitted when the configuration for a data plane `>= 3.1.0` has `http_response_header_for_traceid` created + set to the default value (due to the request coming from a control plane `< 3.1.0`)?

mikefero: I saw the addition, but missed what it was doing; sorry for the confusion.

mikefero (approved):

PR #514, opened 2022-11-16
feat: add v2 expected config hash endpoint

mikefero:
This feature adds created_at and updated_at fields to the GetHashResponse message while ensuring the older v1 endpoint functionality remains intact; however deprecated. These additional fields can be used in combination with the node status fields to get a more complete picture of the syncing process between the control plane and its connected data planes.

#### v1

```json
{
    "expected_hash": "1f74d9c7023b53991f74d9c7023b5399"
}
```

#### v2

```json
{
    "item": {
        "created_at": 1668562104,
        "expected_hash": "1f74d9c7023b53991f74d9c7023b5399",
        "updated_at": 1668633461
    }
}
```


internal/server/admin/handler.go
```diff
@@ -323,7 +339,7 @@ func RegisterAdminService(server *grpc.Server, opts HandlerOpts) {
 	v1.RegisterTargetServiceServer(server, services.target)
 	v1.RegisterSchemasServiceServer(server, services.schemas)
 	v1.RegisterNodeServiceServer(server, services.node)
-	v1.RegisterStatusServiceServer(server, services.status)
+	v1.RegisterStatusServiceServer(server, services.statusV1)
```
GGabriele, line 342: Don't we need to register `services.statusV2` as well?

mikefero: Good catch!

mikefero: Small change request so I force pushed the change.

GGabriele (approved):

PR #515, opened 2022-11-17
feat: custom error message and code for rate-limiting

omegabytes:
Adds user-configurable error_code and error_message fields to the rate-limiting plugin for v3.1^.

### DIFF
--- a/internal/test/e2e/plugins.go
+++ b/internal/test/e2e/plugins.go
@@ -279,6 +279,8 @@ var VersionCompatibilityOSSPluginConfigurationTests = []VersionCompatibilityPlug
 		ConfigureForService: true,
 		ConfigureForRoute:   true,
 	},
+	// DP < 3.1:
+	//   - remove 'error_code', 'error_message'
 	{
 		Name: "rate-limiting",
 		Config: `{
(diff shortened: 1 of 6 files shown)


aboudreault:
ref: https://github.com/Kong/kong/pull/8930

internal/test/e2e/plugins.go
```diff
@@ -279,6 +279,8 @@ var VersionCompatibilityOSSPluginConfigurationTests = []VersionCompatibilityPlug
 		ConfigureForService: true,
 		ConfigureForRoute:   true,
 	},
+	// DP < 3.1:
```
GGabriele, line 282: While you are at it, would you mind documenting the expected changes for older DPs too? Like that `redis_*` fields are removed for DP < 2.7 etc

aboudreault:
LGTM Alex, let's wait for another approval to make sure I am not missing something.

internal/test/e2e/plugins.go
omegabytes: [cfa8dd2](https://github.com/Kong/koko/pull/515/commits/cfa8dd288786e9cb4951be2d4f6a3e007e9e5a62)

Also noticed that `redis_password` is not being removed? https://github.com/Kong/koko/blob/69228016c6c464471b8ea53bb5426677e073f2ff/internal/server/kong/ws/config/compat/compatibility_table.go#L382-L398

aboudreault: that seems correct. `redis_password` was added in a previous kong version. ref: https://github.com/Kong/kong/pull/8032

GGabriele (approved):

PR #532, opened 2022-11-29
Fix/aws lambda error

GGabriele:
Ref: https://github.com/Kong/kong/commit/819d31634fa3b1af06db8a8ebd4c86214de7db8b

### DIFF
--- a/internal/server/kong/ws/config/compat/compatibility_table.go
+++ b/internal/server/kong/ws/config/compat/compatibility_table.go
@@ -570,6 +570,11 @@ var (
 				RemoveFields: []string{
 					"proxy_scheme",
 				},
+				DisableChangeTracking: func(rawJSON string) bool {
+					plugin := gjson.Parse(rawJSON)
+					// do not emit change if 'proxy_scheme' is not set
+					return config.ValueIsEmpty(plugin.Get("config.proxy_scheme"))
+				},
 			},
 		},
 		{
--- a/internal/server/kong/ws/config/compat/compatibility_table.go
+++ b/internal/server/kong/ws/config/compat/compatibility_table.go
@@ -980,6 +980,41 @@ var (
 				},
 			},
 		},
+		{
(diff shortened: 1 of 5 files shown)


internal/server/kong/ws/config/compat/compatibility_table.go
```diff
@@ -570,6 +570,11 @@ var (
 				RemoveFields: []string{
 					"proxy_scheme",
 				},
+				DisableChangeTracking: func(rawJSON string) bool {
+					plugin := gjson.Parse(rawJSON)
+					// do not emit change if 'proxy_scheme' is not set
+					return config.ValueIsEmpty(plugin.Get("config.proxy_scheme"))
```
aboudreault, line 576: :+1:

internal/test/e2e/plugins.go
```diff
@@ -72,12 +72,15 @@ var VersionCompatibilityOSSPluginConfigurationTests = []VersionCompatibilityPlug
...
+	//   - remove 'aws_assume_role_arn' and 'aws_role_session_name'
 	{
 		Name: "aws-lambda",
 		Config: `{
 			"aws_region": "AWS_REGION",
 			"host": "192.168.1.1",
-			"function_name": "FUNCTION_NAME"
+			"function_name": "FUNCTION_NAME",
+			"aws_assume_role_arn": "foo",
+			"aws_role_session_name": "kong"
```
aboudreault, line 83: in some PRs, we added also tests for a 3.0 config and make sure fields were passed (but I also see we don´t do it in all PRs) Should we follow this and add an aws plugin config with these fields in a `...After3.0` test?

GGabriele: Good call. Added it here https://github.com/Kong/koko/pull/532/commits/f51b41f1e15b0282e989831083274e16843b9267

aboudreault (approved):

PR #533, opened 2022-11-29
fix: correct format of Vaults' compatibility message

GGabriele:


internal/server/kong/ws/config/compat/compatibility_table.go
```diff
@@ -45,7 +45,7 @@ func standardPluginNotAvailableMessage(pluginName string, versionWithFeatureSupp
 
 func standardCoreEntityMessage(entityName string, versionWithFeatureSupport string) string {
 	return fmt.Sprintf("The '%s' entity is being used, "+
-		"but Kong Gateway versions < %s do not support this entity. "+
+		"but Kong Gateway versions < %s do not support this entity. ",
```
hbagdi, line 48: Why is there no test?

GGabriele: We currently have no tests for these kind of messages. This is definitely something we need to track.

PR #534, opened 2022-11-29
feat: add k8s config fields to HCV vault

GGabriele:


aboudreault (commented):
I was expecting to see a schema change that validate that auth_method is one of token or kubernetes, should this be added?

aboudreault:
ah I see. it is only for HCV.. but we need to add the main struct stuff here.

aboudreault (approved):

internal/grpc/proto/kong/admin/model/v1/vault.proto
```diff
@@ -23,6 +23,9 @@ message Vault {
     string mount = 4;
     string kv = 5;
     string token = 6;
+    string auth_method = 7;
```
tjasko, line 26: Hmm, realized we're seemingly missing validation for many of these vault config fields... but not required to cover in this PR, just wanted to point it out.

tjasko (approved):

PR #535, opened 2022-11-29
fix: make service tls_verify a nullable boolean

aboudreault:
Currently, the admin API only returns the service `tls_verify` if set to true. If the field is not present or set to false, it is considered as the default value and it is not stored in DB, thus never returned in http responses. This is problematic because that field have 3 meanings:

`true`: tls verification should happen
`false`: tls verification should not happen
`null`/`not specified`: tls verification decision depends on the DP configuration.

This PR makes that field a nullable boolean, so it is always included in the object definition if set to true or false.

#### Before
**tls_verify not set**
tls_verify not stored in the DB
tls_verify not returned in the admin http response

**tls_verify set to null**
tls_verify not stored in the DB
tls_verify not returned in the admin http response

**tls_verify set to false**
tls_verify not stored in the DB
tls_verify not returned in the admin http response

**tls_verify set to true**
tls_verify stored in the DB
tls_verify returned in the admin http response

#### After
**tls_verify not set**
tls_verify not stored in the DB
tls_verify not returned in the admin http response

**tls_verify set to null**
tls_verify not stored in the DB
tls_verify not returned in the admin http response

**tls_verify set to false**
tls_verify stored in the DB
tls_verify returned in the admin http response

**tls_verify set to true**
tls_verify stored in the DB
tls_verify returned in the admin http response


aboudreault:
not sure this solution will work as is unfortunately, https://github.com/Kong/koko/actions/runs/3576982648/jobs/6015440079#step:8:10

javierguerragiraldez:
>not sure this solution will work as is unfortunately

unfortunately, that's not possible in protobuf3.  by design, there's no distinction between the default value (false for booleans) and unset.  specifically, the encoding will just not encode boolean false, number zero, empty strings/tables, etc.

aboudreault:
@javier yeah, in fact that why I modified the proto spec to be a pointer to a boolean. It is working. the part I'm not sure is about if we can break the current protobuf spec of not.

javierguerragiraldez:
> I'm not sure is about if we can break the current protobuf spec or not.

i don't think we publish the grpc API.  as long as it better expresses the HTTP API, it should be ok.  judging from the admin api tests, it seems the gateway code will handle the same expected semantics (that is, differentiating between true/false/absent values in JSON).

if the only issue is the lint complaining that the grpc schema has changed, i'd say we should ignore it.

aboudreault:
@javierguerragiraldez thanks for your feedback. I wasn't sure who was the grpc api consumers. And yes, the admin API handles that correctly and this change is not affecting anything with the existing data in the DB.

javierguerragiraldez (dismissed):

aboudreault:
I fix a minor thing in tests. Javier's approval was dismissed. Looking for another approval to merge.

internal/resource/service_test.go
```diff
@@ -52,6 +53,8 @@ func TestService_ProcessDefaults(t *testing.T) {
 		r.Service.Retries = 1
 		r.Service.Protocol = "grpc"
 		r.Service.Enabled = wrapperspb.Bool(false)
+		tlsVerify := false
+		r.Service.TlsVerify = &tlsVerify
```
GGabriele, line 57: Just fyi: you could already use `r.Service.TlsVerify = falsy`

GGabriele (approved):
Nice!!!

internal/resource/service_test.go
aboudreault: interesting, getting `undefined: falsy` error here?

PR #544, opened 2022-12-07
feat:jwt-parsing utilities

rajkong:
Adding JWT parsing capability so that jwt tokens can be parsed


javierguerragiraldez:
nice.  i guess this could be used to validate keys under the #508 PR.  is there some other use planned? perhaps so koko itself could filter it's own API?

internal/server/util/jwt_service.go
```diff
@@ -0,0 +1,208 @@
...
+	JwtPublicKey string
+}
+
+func New(opts NewJwtServiceOpts) (*JwtService, error) {
+	// parsing in case of bad env variable parsing
+	parsedStr := strings.ReplaceAll(opts.JwtPublicKey, "\\n", "\n")
+
+	raw, rest := pem.Decode([]byte(parsedStr))
+	if len(rest) > 0 {
+		return nil, errors.New("rest found, Jwt key invalid")
```
GGabriele, line 44: The meaning of `rest found` may be clear for those who know what the code is doing, but otherwise it may be a bit obscure. Can we rephrase it more clearly?

internal/server/util/jwt_service.go
```diff
@@ -0,0 +1,208 @@
...
+
+func (s *JwtService) JwtPublicKey() interface{} {
+	return s.jwtPublicKey
+}
+
+var errNoClaim = errors.New("unable to parse claims")
+
+func (s *JwtService) ParseAuthorization(tokenString string) (*JWTClaims, error) {
+	tokenSlice := strings.Split(tokenString, " ")
+	if len(tokenSlice) < 2 { //nolint: gomnd
```
GGabriele, line 63:
```suggestion
	if len(tokenSlice) < 2  || tokenSlice[0] != "Bearer" { //nolint: gomnd
```

Could also add the other check to make the code more concise.

internal/server/util/jwt_service.go
```diff
@@ -0,0 +1,208 @@
...
+		}
+		return jwtClaims, nil
+	}
+	return nil, errNoClaim
+}
+
+// ParseUnverifiedAuthorization Do not use this unless you know what you are doing.
+func ParseUnverifiedAuthorization(tokenString string) (*JWTClaims, error) {
+	tokenSlice := strings.Split(tokenString, " ")
+	if len(tokenSlice) < 2 { //nolint: gomnd
```
GGabriele, line 93: Same as [this](https://github.com/Kong/koko/pull/544/files#r1044176199)

internal/server/util/jwt_service.go
```diff
@@ -0,0 +1,208 @@
+package util
```
GGabriele, line 1: Not sure whether `jwt_service.go` is a good name for this: a `_service.go` in Koko generally represents an grpc/wrpc/rest API service, while this is some more of an util service. What do you think?

internal/server/util/jwt_service.go
rajkong: Done

internal/server/util/jwt_service.go
rajkong: Done

internal/server/util/jwt_service.go
rajkong: Done

internal/server/util/jwt_service.go
rajkong: Done

internal/server/util/helpers.go
```diff
@@ -98,6 +102,28 @@ func LoggerWithSpan(ctx context.Context, l *zap.Logger) *zap.Logger {
 	return l
 }
 
+func LoggerWithAuthElements(authorization string, l *zap.Logger) *zap.Logger {
+	if len(authorization) > 0 {
```
GGabriele, line 106: nit: you can avoid some nesting by doing something like this

```
if len(authorization) <= 0 {
    return l
}
// your logic
}
```
```

GGabriele (approved):
Just a nit. Consider this approved already though.

PR #545, opened 2022-12-09
feat: return more info from Certs APIs

GGabriele:
Sample call/response:

```
$ http :3000/v1/certificates\?cluster.id=4168295f-015e-4190-837e-0fcc5d72a52f cert=$newcert key=$newkey
HTTP/1.1 201 Created
Content-Type: application/json
Date: Fri, 09 Dec 2022 17:41:09 GMT
Transfer-Encoding: chunked

{
    "item": {
        "cert": "-----BEGIN CERTIFICATE-----\n cert \n-----END CERTIFICATE-----",
        "created_at": 1670607669,
        "id": "1faa526c-8920-4ad6-acf5-3588a1474d37",
        "key": "-----BEGIN PRIVATE KEY-----\n key \n-----END PRIVATE KEY-----",
        "metadata": {
            "expiry": 1985967654,
            "issuer": "CN=Yolo42 self-signed testing CA,O=Yolo42\\, Inc.,ST=CA,C=US",
            "snis": [
                "new.com",
                "test.com"
            ],
            "subject": "CN=koko_test,O=Yolo42 Inc.,L=San Francisco,ST=CA,C=US"
        "updated_at": 1670607669
    }
}
```


internal/grpc/proto/kong/admin/model/v1/certificate.proto
```diff
@@ -37,4 +37,9 @@ message Certificate {
     (referenceable) = true
   ];
   repeated string tags = 8;
+  string subject = 9;
+  string issuer = 10;
+  string san = 11;
+  string expiry = 12;
+  string key_usage = 13;
```
tjasko, line 44: We should also log out the extension key usages:
https://pkg.go.dev/crypto/x509#ExtKeyUsage

We can keep it all under the same field, I see no problem combining.

internal/server/admin/certificate_service.go
```diff
@@ -35,6 +82,11 @@ func (s *CertificateService) GetCertificate(ctx context.Context,
 	if err != nil {
 		return nil, s.err(ctx, err)
 	}
+	if !util.IsReference(result.Certificate.Cert) {
+		if err := extractCertInfo(result.Certificate); err != nil {
+			return nil, s.err(ctx, err)
```
tjasko, line 87: No need to use `s.err()` when calling `extractCertInfo()` as it won't be outputting any public-facing error anyway. Same goes for the `certificatesFromObjects()` call too.

internal/server/admin/certificate_service.go
```diff
@@ -35,6 +82,11 @@ func (s *CertificateService) GetCertificate(ctx context.Context,
 	if err != nil {
 		return nil, s.err(ctx, err)
 	}
+	if !util.IsReference(result.Certificate.Cert) {
+		if err := extractCertInfo(result.Certificate); err != nil {
```
tjasko, line 86: Instead of repeating this logic over and over, let's just implement the [model.ObjectWithResourceDTO](https://github.com/Kong/koko/blob/v0.1.0/internal/model/object.go#L43) interface, have it use the default `MarshalResourceJSON()` behavior, and implement its own `UnmarshalResourceJSON()` method that uses the default logic, but handles this cert extraction.

This way, the implementation will be clean and we don't need to litter the service-level code with this.

internal/server/util/helpers.go
```diff
@@ -241,3 +245,8 @@ func grpcRecoveryHandler(logger *zap.Logger) grpcRecovery.Option {
 		return status.Errorf(codes.Internal, "internal server error")
 	})
 }
+
+func IsReference(v string) bool {
```
tjasko, line 249: Can we throw a comment here & likely moving the link below to it?

internal/server/admin/ca_certificate_test.go
```diff
@@ -182,6 +182,81 @@ func TestCACertificateCreate(t *testing.T) {
...
+			Subject: "XXX",
+		}
+		res := c.POST("/v1/ca-certificates").WithJSON(cert).Expect()
+		res.Status(http.StatusBadRequest)
+		body := res.JSON().Object()
+		body.ValueEqual("message", "validation error")
+		body.Value("details").Array().Length().Equal(1)
+		err := body.Value("details").Array().Element(0).Object()
+		err.ValueEqual("type", v1.ErrorType_ERROR_TYPE_ENTITY.String())
+		err.Value("messages").Array().Element(0).String().
```
tjasko, line 197: nit: Should be checking the length of `$.details.messages` too (this comment is valid for most of these tests).

internal/grpc/proto/kong/admin/model/v1/certificate.proto
tjasko, line 44: Can we actually move these fields into its own Protobuf message/entity, that way it can be re-used? I would just call it `CertificateMetadata`.

internal/server/admin/certificate_service.go
GGabriele: Very good point, I forgot we have such tooling already in place!

I tried to implement this, but it's not super clear to me how I'm expected to do so. I added something like the following, but I think I'm missing something since the receiving `r.Certificate` is always empty:

```

func (r Certificate) MarshalResourceJSON() ([]byte, error) {
	return json.Marshal(r)
}

func (r Certificate) UnmarshalResourceJSON([]byte) error {
	if !isReference(r.Certificate.Cert) {
		if err := extractCertInfo(r.Certificate); err != nil {
			return err
		}
	}
	return json.Unmarshal([]byte(fmt.Sprintf("%v", r)), &v1.Certificate{})
}
```

mikefero (approved):
LGTM. If possible let's add `san_names` and `key_usage` validation to this as well; granted this will require new certs.

mikefero (approved):

GGabriele:
@tjasko are you happy with the changes here? Since we need to revisit this implementation anyway to store the data in the database, do you think we can live with the current implementation (no [model.ObjectWithResourceDTO](https://github.com/Kong/koko/blob/v0.1.0/internal/model/object.go#L43)) to ship this quickly?

internal/server/admin/certificate_service.go
tjasko: You'll need to call unmarshal first and then extract the info on the certificate. 😄

internal/grpc/proto/kong/admin/model/v1/certificate.proto
```diff
@@ -37,4 +37,42 @@ message Certificate {
...
   repeated string tags = 8;
+  CertificateMetadata metadata = 9;
+}
+
+message CertificateMetadata {
+  string subject = 1;
+  string issuer = 2;
+  repeated string san_names = 3;
+  int32 expiry = 4;
+  repeated KeyUsageType key_usage = 5;
```
tjasko, line 48: Should this be called `key_usages` now?

internal/server/admin/certificate_service.go
tjasko: Can we ensure that's fixed elsewhere too? Obviously, this comment will be moot once [this](https://github.com/Kong/koko/pull/545#discussion_r1046395371) is done though.

tjasko (commented):
Other than that one comment to hook the certificate metadata extraction into the JSON unmarshal'ing, everything else looks great. Nice work @GGabriele! 😄

internal/resource/ca_certificate.go
```diff
@@ -159,6 +162,39 @@ func (r CACertificate) Indexes() []model.Index {
 	}
 }
 
+// MarshalResourceJSON is used here to make sure that CA certificate
```
tjasko, line 165: nit: May want to make a comment about not saving the certificate metadata in the DB for now, but it's obvious by the code, so works with me.

internal/server/admin/certificate_service.go
```diff
@@ -222,7 +226,26 @@ func (s *CertificateService) logger(ctx context.Context) *zap.Logger {
...
+			"retrieving snis used by certificate '%s': %w",
+			cert.Id, err,
+		)
+	}
+	if len(snis) > 0 && cert.Metadata == nil {
+		cert.Metadata = &pb.CertificateMetadata{}
+	}
+	for _, sni := range snis {
+		cert.Metadata.Snis = append(cert.Metadata.Snis, sni.Name)
+	}
```
tjasko, line 150: nit: This can just be:
```suggestion
	if len(snis) > 0 && cert.Metadata == nil {
		cert.Metadata = &pb.CertificateMetadata{Snis: make([]string, len(snis))}
	}
	for i, sni := range snis {
		cert.Metadata.Snis[i] = sni.Name
	}
```

internal/server/admin/certificate_service.go
```diff
@@ -235,7 +258,49 @@ func certificatesFromObjects(objects []model.Object) ([]*pb.Certificate, error)
 				return nil, err
 			}
 		}
+		// retrieve the snis currently using this certificate
+		if err := addSNIMetadata(ctx, cert, db); err != nil {
```
tjasko, line 165: Ideally, this would happen within the same transaction to adhere to ACID, but that's easier said than done right now. This isn't likely to be a problem for most, but figured to mention it as it's something we should try to iterate on.

Nothing needs to be done for this right now though.

tjasko (approved):
Looks great! Some nits, but nothing blocking rolling this out. Great work @GGabriele!

PR #551, opened 2022-12-13
feat: introducing customizable model object cascading functionality

tjasko:
This introduces a new interface that objects can implement in order to influence its behavior within the persistence store. The primary use-case for this right now is to override the object cascading functionality upon deletions.

This change is primarily being introduced for the ongoing consumer groups effort, though this functionality may come in handy later on.

### DIFF
--- a/internal/model/registry.go
+++ b/internal/model/registry.go
@@ -87,3 +87,15 @@ func ObjectFromProto(p proto.Message) (Object, error) {
 	// match any underlining resource on our model objects.
 	return nil, fmt.Errorf("cannot find type from Protobuf message %s", expectedFullName)
 }
+
+// OptionsForType returns the options for the provided Type, granted that options exist
+// for the given type. In the event no options exist, DefaultObjectOptions is returned.
+func OptionsForType(typ Type) ObjectOptions {
+	if obj, err := NewObject(typ); err == nil {
+		if obj, ok := obj.(ObjectWithOptions); ok {
+			return obj.Options()
+		}
+	}
+
+	return DefaultObjectOptions
+}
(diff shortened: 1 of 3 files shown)


internal/store/index.go
```diff
@@ -221,10 +221,28 @@ func (s *ObjectStore) onDeleteCascade(ctx context.Context,
...
-			err := s.delete(ctx, tx, model.Type(typ), id)
-			if err != nil {
+
+			// Handle deletion of the foreign key relation, but skip cascading
+			// the entire object whenever applicable.
+			//
+			// This is useful for one-to-many relationships, like what is used
+			// for consumer groups. As when a consumer is deleted, we'll want
+			// to delete the association, but not the consumer group itself.
+			if !model.OptionsForType(typ).CascadeOnDelete {
```
tjasko, line 233: Considered implementing unit tests for this, but this part of the code base is very much so lacking them. I can write them if we must, but do what to say that this will be tested indirectly by the consumer group integration tests.

internal/model/registry.go
```diff
@@ -87,3 +87,15 @@ func ObjectFromProto(p proto.Message) (Object, error) {
 	// match any underlining resource on our model objects.
 	return nil, fmt.Errorf("cannot find type from Protobuf message %s", expectedFullName)
 }
+
+// OptionsForType returns the options for the provided Type, granted that options exist
+// for the given type. In the event no options exist, DefaultObjectOptions is returned.
+func OptionsForType(typ Type) ObjectOptions {
+	if obj, err := NewObject(typ); err == nil {
```
tjasko, line 94: Would normally test this, but it'd require adding in a dummy type to be added and this code is not written in a way to remove the global state that's currently going on. There's some refactoring that really should happen to avoid this global state.

GGabriele (approved):

rajkong (approved):
LGTM

PR #552, opened 2022-12-13
feat: allowing persistence store to list foreign objects

tjasko:
This introduces a change to the persistence store that allows callers to list the foriegn resources themselves. Currently, a `ListFor()` call exists that allows listing resource(s) that have the associated foreign key index, although listing the other way around was not previously supported.

### DIFF
--- a/internal/store/opts.go
+++ b/internal/store/opts.go
@@ -73,10 +73,12 @@ func DeleteByType(typ model.Type) DeleteOptsFunc {
 }
 
 type ListOpts struct {
-	ReferenceType model.Type
-	ReferenceID   string
-	PageSize      int
-	Page          int
+	ReferenceType          model.Type
+	ReferenceID            string
+	ReferenceReverseLookup bool
+
+	PageSize int
+	Page     int
 
 	// CEL expression used for filtering.
 	// Read more: https://github.com/google/cel-spec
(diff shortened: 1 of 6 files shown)


internal/store/opts.go
```diff
@@ -124,13 +126,52 @@ func NewListOpts(fns ...ListOptsFunc) (*ListOpts, error) {
...
+//     service: {id: service-1}
+//   - id: route-2
+//     service: {id: service-1}
+//   - id: route-3
+//
+// When calling `ListFor(resource.TypeRoute, "route-1")` with
+// `resource.NewList(resource.TypeService)`.
+//
+// Then this would return the service with ID `service-1`.
+func ListReverseFor(typ model.Type, id string) ListOptsFunc {
```
tjasko, line 167: I'm not sure what else to call this, but do you feel it's confusing? At least the commentary clears up when to use `ListFor()` vs `ListReverseFor()`. `ListFor()` wasn't named very well in the first place, so that's part of the confusion.

internal/store/opts.go
```diff
@@ -124,13 +126,52 @@ func NewListOpts(fns ...ListOptsFunc) (*ListOpts, error) {
...
+// As such, this is used to list the foreign resources themselves.
+//
+// Given the following routes:
+//   - id: route-1
+//     service: {id: service-1}
+//   - id: route-2
+//     service: {id: service-1}
+//   - id: route-3
+//
+// When calling `ListFor(resource.TypeRoute, "route-1")` with
```
GGabriele, line 163:
```suggestion
// When calling `ListReverseFor(resource.TypeRoute, "route-1")` with
```

Should be this?

GGabriele (approved):
Awesome job!

internal/store/store.go
```diff
@@ -443,8 +444,20 @@ func (s *ObjectStore) referencedList(ctx context.Context, list model.ObjectList,
...
+		// When `opt.ReferenceReverseLookup` is not true, we're extracting <id> from the below key:
+		// `c/.../ix/f/<typ>/<opt.ReferenceID>/<opt.ReferenceType>/<id>`
+		//
+		// When it is true, we're extracting the <id> from the below key:
+		// `c/.../ix/f/<typ>/<id>/<opt.ReferenceType>/<opt.ReferenceID>`
+		id := key[keyPrefixLen:]
+		if opt.ReferenceReverseLookup {
+			startIdx := strings.Index(keyPrefix, persistence.WildcardOperator)
+			endIdx := strings.Index(key[startIdx:], "/")
+			id = key[startIdx : startIdx+endIdx]
```
javierguerragiraldez, line 457: seems a bit fragile.  i think `strings.Cut` has better behaviour when the separator isn't found.

```suggestion
			startIdx := strings.Index(keyPrefix, persistence.WildcardOperator)
			if startIdx < 0 {
				return fmt.Errorf(.....)
			}
			id, _, _ = strings.Cut(key[startIdx:], "/")
```

internal/store/opts.go
tjasko: Yup, it should be, bad copy pasta. Updated!

internal/store/store.go
tjasko: Fair point, updated 😀

tjasko:
> @javierguerragiraldez please add some tests, this is a new feature with a not-too-obvious requirement (the sql processing downstream). it's easy to imagine unrelated work breaking this if not tested.

As mentioned previously, adding unit tests here is far from trivial given how the persistence store was written. This is indirectly tested, but via the integration tests of the consumer group work. I will agree with you though, we need to get this codebase up to snuff & mocking the persistence layer to test this code... it's just difficult due to all the global state happening. Need to de-couple it and be able to register new types on the fly without affecting the state of everything else.

GGabriele (approved):

aboudreault (approved):

PR #553, opened 2022-12-13
feat: add consumer-group definition

GGabriele:


internal/grpc/proto/kong/admin/service/v1/consumer_group.proto
```diff
@@ -0,0 +1,189 @@
...
+  }
+  rpc ListConsumerGroups(ListConsumerGroupsRequest) returns (ListConsumerGroupsResponse) {
+    option (google.api.http) = {get: "/v1/consumer-groups"};
+  }
+
+  rpc ListConsumerGroupMembers(ListConsumerGroupMembersRequest) returns (ListConsumerGroupMembersResponse) {
+    option (google.api.http) = {get: "/v1/consumer-groups/{id}/members"};
+  }
+  rpc CreateConsumerGroupMember(CreateConsumerGroupMemberRequest) returns (CreateConsumerGroupMemberResponse) {
+    option (google.api.http) = {post: "/v1/consumers/{consumer_id}/groups/{consumer_group_id}/members"};
```
javierguerragiraldez, line 41: what about naming this `AddConsumerToGroup`? also, a POST without body feels unnatural.
maybe both PUT `/v1/consumer-groups/{consumer_group_id}/members/{consumer_id}` and a PUT `/v1/consumers/{consumer_id}/groups/{consumer_group_id}` to the same grpc method.

and both GET `/v1/consumer-groups/{consumer_group_id}/members` => `ListConsumerGroupMembers` and GET `/v1/consumers/{consumer_id}/groups` => `ListConsumerGroupsByConsumer`

i think these names and paths come from the idea of a link index (necessary on relational DB to model a many:many relationship), but for the user it should just be that a consumer has a list of groups it belongs to, and the group has a list of consumers.  the link object shouldn't be visible.

internal/grpc/proto/kong/admin/model/v1/consumer_group.proto
```diff
@@ -0,0 +1,32 @@
+syntax = "proto3";
+
+package kong.admin.model.v1;
+
+import "google/api/field_behavior.proto";
+
+option go_package = "github.com/kong/koko/internal/gen/grpc/kong/admin/model/v1;v1";
+
+message Member {
```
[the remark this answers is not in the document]
GGabriele: Errr, that's a leftover. Good catch!

Fixed.

internal/grpc/proto/kong/admin/service/v1/consumer_group.proto
tjasko: It's generally best to stick to the standard method naming conventions, and in this case, `CreateConsumerGroupMember` adheres to that:
https://google.aip.dev/136

POST requests are not expected to be idempotent, PUT requests are. Because the same member cannot be added again, a POST request is technically proper here given standard RESTful API design patterns.

The `ListConsumerGroupsByConsumer` renaming is totally valid though, have no problem renaming that as it is more specific.

GGabriele: We haven't implemented yet GET on `/v1/consumers/{consumer_id}/groups`, so I think that `ListConsumerGroupsByConsumer`, which is a totally valid name, is not necessary at the moment.

aboudreault (dismissed):

PR #554, opened 2022-12-13
fix: make unique key check concurrent safe

aboudreault:
** Work In Progress **

TODO:

- [x] fix postgres unit tests
- [x] fix mysql

### DIFF
--- a/internal/persistence/error.go
+++ b/internal/persistence/error.go
@@ -8,6 +8,9 @@ import (
 // expected amount. This error should usually not be seen under normal circumstances.
 var ErrInvalidRowsAffected = fmt.Errorf("invalid rows affected")
 
+// ErrUniqueViolation is used to indicate a unique constraint violation error.
+var ErrUniqueViolation = fmt.Errorf("unique constraint violation error")
+
 // ErrNotFound is used to indicate the key being looked up was not found in the datastore.
 type ErrNotFound struct {
 	Key string
--- a/internal/persistence/mysql/mysql.go
+++ b/internal/persistence/mysql/mysql.go
@@ -8,6 +8,7 @@ import (
 	"time"
 
 	sq "github.com/Masterminds/squirrel"
+	"github.com/go-sql-driver/mysql"
 	"github.com/google/cel-go/common/operators"
 	"github.com/kong/koko/internal/json"
 	"github.com/kong/koko/internal/persistence"
@@ -16,7 +17,11 @@ import (
 )
 
 // ErrMariaDBUnsupported is the error returned when db.DialectMariaDB is attempted to be used.
-var ErrMariaDBUnsupported = errors.New("MariaDB is currently unsupported")
+var (
+	ErrMariaDBUnsupported = errors.New("MariaDB is currently unsupported")
+)
+
+const duplicateEntryErrorCode uint16 = 1062
 
 // MySQL defines a persistence store integration for databases that speak the MySQL protocol.
 //
@@ -51,6 +56,10 @@ func (s *MySQL) SetDefaultSQLOptions(db *sql.DB) error {
 
 func (s *MySQL) Get(ctx context.Context, key string) ([]byte, error) {
 	return (&mysqlQuery{s.readOnlyDB, s.queryTimeout}).Get(ctx, key)
+}
+
+func (s *MySQL) Insert(ctx context.Context, key string, value []byte) error {
+	return (&mysqlQuery{s.db, s.queryTimeout}).Insert(ctx, key, value)
 }
 
 func (s *MySQL) Put(ctx context.Context, key string, value []byte) error {
@@ -159,6 +168,28 @@ func (t *mysqlQuery) Get(ctx context.Context, key string) ([]byte, error) {
 	}
 
 	return value, nil
+}
+
+func (t *mysqlQuery) Insert(ctx context.Context, key string, value []byte) error {
+	rawSQL, placeholders, err := sq.StatementBuilder.
+		Insert("store").
+		Columns("`key`", "value").
+		Values(key, value).
+		ToSql()
+	if err != nil {
+		return err
+	}
+	ctx, cancel := context.WithTimeout(ctx, t.queryTimeout)
+	defer cancel()
+	if _, err = t.ExecContext(ctx, rawSQL, placeholders...); err != nil {
+		if mySQLErr, ok := err.(*mysql.MySQLError); ok {
+			if mySQLErr.Number == duplicateEntryErrorCode {
+				return persistence.ErrUniqueViolation
+			}
+		}
+		return err
+	}
+	return nil
 }
 
 func (t *mysqlQuery) Put(ctx context.Context, key string, value []byte) error {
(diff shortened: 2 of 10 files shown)


internal/persistence/error.go
```diff
@@ -8,6 +8,9 @@ import (
 // expected amount. This error should usually not be seen under normal circumstances.
 var ErrInvalidRowsAffected = fmt.Errorf("invalid rows affected")
 
+// ErrUniqueViolation is used to indicate a data constraint error.
+var ErrUniqueViolation = fmt.Errorf("unique violation error")
```
rajkong, line 12: nit: "unique constraint violation error"

rajkong (commented):
Overall looks ok. I think it would be good to make sure that the upsert test cases work without any impact

aboudreault:
I changed the approach a little bit.. rather than making `Put` failing  if a key exists, I introduced a new `Insert` function. Insert will raise `ErrUniqueViolation` on error. This is less confusing and also still allow cases that we are fine with multiple concurrent updates to work (like event updates). Currently, `Insert` is used by `store.Create` and `createIndexes` (which is fine because Upsert is deleted old indexes and recreate them)

GGabriele:
> I changed the approach a little bit.. rather than making `Put` failing if a key exists, I introduced a new `Insert` function. Insert will raise `ErrUniqueViolation` on error. This is less confusing and also still allow cases that we are fine with multiple concurrent updates to work (like last event updates). Currently, `Insert` is used by `store.Create` and `createIndexes` (which is fine because Upsert is deleting old indexes and recreate them)

That makes sense, I was a bit confused from the first iteration, thanks!

Can you add a commit message describing in more details what this change does? It's still not super clear to me how [this](https://github.com/Kong/koko/blob/7dd1c7b31b7b146b738f71613859855f23659237/internal/persistence/postgres/postgres.go#L21) is thread safe while [this](https://github.com/Kong/koko/blob/7dd1c7b31b7b146b738f71613859855f23659237/internal/persistence/postgres/postgres.go#L22) wasn't. Does the `on conflict` clause change the isolation level?

aboudreault:
@GGabriele good point, I will add the explanation in the commit message.

aboudreault:
@GGabriele to answer you question, `on conflict` doesn't change anything about the isolation level. Although the initial problem is that indexes existence are checked on the same db snapshot when multiple transactions are created at the same time, the primary key unique violation constraint of the store table should still happen when committing. This is not the case due to the `on conflict` which is basically just causing `lost updates` and overwrite the previous data.

aboudreault:
The serializable isolation level can help this kind of cases too, because it doesn't allow multiple concurrent updates modifying the same row. But in our specific case, the main primary key will do the job.

GGabriele (approved):
Thanks a lot for investigating and fixing this. Awesome job!!!

internal/persistence/mysql/mysql.go
```diff
@@ -16,7 +17,11 @@ import (
 )
 
 // ErrMariaDBUnsupported is the error returned when db.DialectMariaDB is attempted to be used.
-var ErrMariaDBUnsupported = errors.New("MariaDB is currently unsupported")
+var (
+	ErrMariaDBUnsupported = errors.New("MariaDB is currently unsupported")
+
+	duplicateEntryErrorCode uint16 = 0x426
```
tjasko, line 23: Why not just write out `duplicateEntryErrorCode uint16 = 1062`? Any reason to specifically use hex?

internal/persistence/error.go
```diff
@@ -8,6 +8,9 @@ import (
 // expected amount. This error should usually not be seen under normal circumstances.
 var ErrInvalidRowsAffected = fmt.Errorf("invalid rows affected")
 
+// ErrUniqueViolation is used to indicate a unique constraint violation error.
+var ErrUniqueViolation = fmt.Errorf("unique constraint violation error")
```
tjasko, line 12: (this part of the comment is basically moot due to the last paragraph)

If this were to occur, this error would not be properly returned to callers. We'll need to update the `store.HandleErr()` to check for this & then re-write the error to a public-facing error.

Additionally, should this be before to the `store` package, as that's where the `store.ErrNotFound` error exists today?

Can we get this covered by an integration test?

----

Lastly, why create a new error for this in the first place? Why aren't we using `store.ErrConstraint` instead?

internal/persistence/mysql/mysql.go
tjasko, line 23: nit: Can we make this a constant instead?

internal/store/index.go
```diff
@@ -93,8 +93,13 @@ func (s *ObjectStore) createIndexes(ctx context.Context,
 				return err
 			}
 
-			err = tx.Put(ctx, key, value)
+			err = tx.Insert(ctx, key, value)
```
tjasko, line 96: I'm not sure this is what we want to do here. If an object is being created (not upserted), this is fine, however in the event it's being upserted, wouldn't this always block the upsert as long as there's a unique index on that object?

This code path is called for both create & upsert.

aboudreault: The Upsert code is already deleting indexes then recreate them. https://github.com/Kong/koko/blob/main/internal/store/store.go#L193

internal/persistence/mysql/mysql.go
aboudreault: +1

aboudreault: not really. that what I had when debugging... will use 1062 for clarity with the mysql docs.

internal/persistence/error.go
aboudreault: Actually, I used the same pattern than we have for the different layers. in the persistence one, errors are mapped to what we find in the `persistence/error.go` file. like `persistence.ErrNotFound`.: https://github.com/Kong/koko/blob/main/internal/persistence/error.go#L12

then, at the higher level (store), they are mapped to the proper `store.Err*` ones: like [here](https://github.com/Kong/koko/blob/main/internal/store/store.go#L302) and [here](https://github.com/Kong/koko/pull/554/files#diff-e229227e6e2d6b98849b281cf747ac752e94e82988e36592273e99528f57cbcaR98)

which is then handled by `HandleErr`

aboudreault: To get this covered this by an integration tests, it will need some work to get multiple concurrent transactions because at the moment, KoKo is already checking itself if the key exists then return ErrConstraint.

internal/store/index.go
tjasko: Ohh derp, right.

GGabriele (approved):

aboudreault:
Thanks for the review team, and also to @rajkong and @hbagdi  for the help and feedback for this issue.

tjasko:
@aboudreault Didn't finish reviewing, but trust you & team this is the fix :)

aboudreault:
@tjasko crap, sorry about that. I saw your last comments then I saw you were doing other stuff in another one and assume it was done :man_facepalming:. Please, feel to comment here, I can address anything if needed.

PR #556, opened 2022-12-14
feat: introducing concept of unmanaged indexes

tjasko:
This allows the ability to bypass the default managed index behavior that the persistence store provides, in order to explicitly set whether index(es) should be created and/or removed.

This allows the ability to add & remove foreign relations without needing to know all foreign relations that exist on the given entity.

### DIFF
--- a/internal/model/object.go
+++ b/internal/model/object.go
@@ -89,6 +89,36 @@ const (
 	IndexForeign TypeIndex = "foreign"
 )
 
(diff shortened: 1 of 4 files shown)


GGabriele (approved):
The detailed doc strings are very much appreciated! Thanks a lot!

GGabriele (dismissed):

GGabriele (approved):

PR #558, opened 2022-12-14
feat: introducing foreign key management support to persistence store

tjasko:
This change implements the ability to update just the foreign keys related to an object. This functionality is required for the recent consumer groups implementation.

### DIFF
--- a/internal/store/store.go
+++ b/internal/store/store.go
@@ -551,12 +551,7 @@ func (s *ObjectStore) updateForeignKeysTx(
 					return err
 				}
 
-				// Ensure the foreign relation is not already created.
-				if err := s.checkIndex(gCtx, tx, index, key); err != nil {
-					return err
-				}
-
-				if err := tx.Put(gCtx, key, value); err != nil {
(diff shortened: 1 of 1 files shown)


internal/store/store.go
```diff
@@ -463,6 +516,75 @@ func (s *ObjectStore) referencedList(ctx context.Context, list model.ObjectList,
...
+				if _, err = tx.Get(gCtx, fk); err != nil {
+					if errors.As(err, &persistence.ErrNotFound{}) {
+						return ErrConstraint{Index: index}
+					}
+					return err
+				}
+
+				if err := tx.Insert(gCtx, key, value); err != nil {
+					return err
+				}
```
aboudreault, line 556: In this case, you can map the ErrUniqueViolation error to a ErrConstraint because you have that information, like the above if statement.

tjasko: Yeah, you're right. Went ahead and updated this.

tjasko: This has been fixed up.

GGabriele (approved):

aboudreault (approved):

PR #559, opened 2022-12-15
feat: add consumer-groups types for version compatibility

GGabriele:


internal/server/kong/ws/config/version_compatibility.go
```diff
@@ -41,14 +41,37 @@ const (
...
+	ConsumerGroup
+	ConsumerGroupConsumer
+	ConsumerGroupPlugin
 )
 
 func (u UpdateType) String() string {
-	return [...]string{"plugin", "plugin", "service", "route", "upstream", "vault"}[u]
+	return [...]string{
+		"plugin",
+		"plugin",
```
aboudreault, line 52: was there before, but why twice "plugin"?

aboudreault (approved):

internal/server/kong/ws/config/version_compatibility.go
GGabriele: honestly, this is a bit silly, but it's because the first and second `UpdateType`s are `Plugin` and `CorePlugin`, which are represented the same way in the config table in the DP. We should definitely clean up/refactor this folder at some point.

aboudreault: ah, thanks for clarification.

PR #563, opened 2022-12-16
fix: use underscore in labels

omegabytes:
Names for metrics & labels should use valid characters for prometheus.
See: https://prometheus.io/docs/concepts/data_model/#metric-names-and-labels

### DIFF
--- a/internal/server/kong/ws/payload.go
+++ b/internal/server/kong/ws/payload.go
@@ -120,12 +120,12 @@ func (p *Payload) configForVersion(version string) (cacheEntry, error) {
 				zap.Int("total-resource-changes", resourcesUpdated))
 			metrics.Gauge("version_compatibility_configuration_change_count",
 				int64(changeCount), metrics.Tag{
-					Key:   "kong-dp-version",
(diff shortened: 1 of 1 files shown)


zhongchen (approved):

internal/server/kong/ws/payload.go
```diff
@@ -107,7 +107,7 @@ func (p *Payload) configForVersion(version string) (cacheEntry, error) {
 		if err != nil {
 			p.logger.Error("failed to process config table update",
 				zap.Error(err),
-				zap.String("kong-dp-version", version),
+				zap.String("kong_dp_version", version),
```
tjasko, line 110: We can leave the Zap labels alone, as we do stick the `kebab-case` there.

omegabytes:
Interestingly, `promlinter` does not fail validation on `-`. It doesn't seem that `-` is automatically converted to `_`, either.

internal/server/kong/ws/payload.go
omegabytes: Fixed

GGabriele (approved):

